PR #4, opened 2023-03-30
Generate secret locally 

bumi:


src/webln/NostrWeblnProvider.ts
```diff
@@ -213,6 +236,40 @@ export class NostrWebLNProvider {
       });
     });
   }
+
+  initNWC(provider: string, options: { name: string, returnTo?: string }) {
+    const height = 600;
+    const width = 400;
+    const top = window.outerHeight / 2 + window.screenY - height / 2;
+    const left = window.outerWidth / 2 + window.screenX - width / 2;
+    // @ts-ignore for my TS love :)
```
rolznz, line 245: you could make NWC_URLS be a `Record<string, string>` or make provider be a `keyof typeof NWC_URLS`. I think the latter is better because this will only work if a provider exists in NWC_URLs. Or, if you add another parameter to options so a user can provide a URL themselves?

rolznz:
Could you add an example to the README for the usage?

src/webln/NostrWeblnProvider.ts
```diff
@@ -213,6 +236,40 @@ export class NostrWebLNProvider {
       });
     });
   }
+
+  initNWC(provider: string, options: { name: string, returnTo?: string }) {
+    const height = 600;
```
rolznz, line 241: I have a feeling this is not going to work well on mobile (not just the responsive part, but launching popups). Is there a reason you didn't go with the callback idea?

rolznz: I guess this is what returnTo is for? in my opinion if it is passed, it should open the url in the current window rather than launching the popup

src/webln/NostrWeblnProvider.ts
```diff
@@ -213,6 +236,40 @@ export class NostrWebLNProvider {
...
+    if (options.returnTo) {
+      url.searchParams.set('returnTo', options.returnTo);
+    }
+    return new Promise((resolve) => {
+      const popup = window.open(
+        url.toString(),
+        `${document.title} - Wallet Connect`,
+        `height=${height},width=${width},top=${top},left=${left}`
+      );
+      let processingResponse = false; // make sure we request the access token only once
```
rolznz, line 259: Would this be needed if you removed the event listener when you get a message?

src/webln/NostrWeblnProvider.ts
```diff
@@ -213,6 +236,40 @@ export class NostrWebLNProvider {
...
   }
+
+  initNWC(provider: string, options: { name: string, returnTo?: string }) {
+    const height = 600;
+    const width = 400;
+    const top = window.outerHeight / 2 + window.screenY - height / 2;
+    const left = window.outerWidth / 2 + window.screenX - width / 2;
+    // @ts-ignore for my TS love :)
+    const url = new URL(NWC_URLS[provider]);
+    url.searchParams.set('c', options.name);
```
rolznz, line 247: `c` is for client right? what if the client wants to generate multiple connection strings? would they just add a suffix or something?

bumi: why would a client want multiple connections? 
the c is just the app name passed to NWC.

src/webln/NostrWeblnProvider.ts
bumi: good point, I did not remove the handler

src/webln/NostrWeblnProvider.ts
bumi: This is for browser contexts.
and this is a helper, if somebody wants it differently they can do it differently. 
most OAuth apps typically open a popup I also like this better than completely navigating the user away from the website.

src/webln/NostrWeblnProvider.ts
bumi: how do I do this? 

I wanted the NWC URLs to be preconfigured, but a custom URL should also be possible, true.

src/webln/NostrWeblnProvider.ts
rolznz: @bumi yeah exactly so you know what session is being used for so that you e.g. don't accidentally delete the wrong session. Another example might be lightsats if we had multiple tips. It might be good to be able to differentiate them somehow, maybe a description field or something?

src/webln/NostrWeblnProvider.ts
rolznz: It seems you already changed it and it's a bit more complicated now since it can be a URL or a provider name. I think the best you can easily do is:

```const urlStr = NWC_URLS[providerNameOrUrl as keyof typeof NWC_URLS] || providerNameOrUrl;```

**Review by rolznz — APPROVED** [review]
@bumi I added a few changes and an example

File: README.md [hunk 5]
```diff
@@ -73,6 +72,22 @@ console.log(response.preimage);
 webln.close(); // close the websocket connection
 ```
 
+#### Generate a new NWC connect url
+```js
+const webln = webln.NostrWebLNProvider.withNewSecret({
+  walletPubkey: process.env.NEXT_PUBLIC_NWC_WALLET_PUBKEY,
+});
```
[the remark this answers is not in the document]
rolznz: @bumi sorry about that 😅 updated.

PR #5, opened 2023-04-13
Fix: NWC subscriptions

rolznz:


src/webln/NostrWeblnProvider.ts
```diff
@@ -159,16 +161,46 @@ export class NostrWebLNProvider {
...
+        {
+          kinds: [23195, 23196],
+          authors: [this.walletPubkey],
+          "#e": [event.id],
+        }
+      ]);
+
+      function replyTimeout() {
+        //console.error(`Reply timeout: event ${event.id} `);
+        reject(`reply timeout: event ${event.id}`);
```
bumi, line 176: we probably should also unsubscribe here to prevent events coming in later.

src/webln/NostrWeblnProvider.ts
```diff
@@ -177,35 +209,12 @@ export class NostrWebLNProvider {
...
-          } else {
-            reject({ error: decryptedContent });
-          }
-        });
       });
     });
   }
+  private checkConnected() {
+    if (!this.connected) {
+      throw new Error("please call enable() and wait for the promise to resolve before calling this function")
```
bumi, line 217:
```suggestion
      throw new Error("please call enable() and await the promise before calling this function")
```

PR #14, opened 2023-05-17
fix: remove cross-fetch

rolznz:
Fixes https://github.com/getAlby/alby-js-sdk/issues/13

I didn't add a check for whether globalThis.fetch exists or not. It is not even guaranteed that globalThis will exist in old browsers and node versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,6 +346,9 @@ console.log(response.keysends);
 }
 ```
 
+## fetch
+This library relies on a global fetch object which will only work in newer browsers and versions of node. You can manually install a global fetch option or polyfill if needed.
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -346,6 +346,9 @@ console.log(response.keysends);
 }
 ```
 
+## fetch
+This library relies on a global fetch object which will only work in newer browsers and versions of node. You can manually install a global fetch option or polyfill if needed.
```
bumi, line 350: we need to give an example here how it is done.

rolznz: @bumi updated

PR #15, opened 2023-05-19
chore: add typings for webhooks and invoices

rolznz:


bumi:
Please ping somebody with more TS knowlege to review. 
Keep in mind that the API response will change. Where is the invoice type used?

rolznz:
@bumi the invoice type is used in the invoice settled webhook post body, and also the response of POST api/invoices.

im-adithya (approved):
utACK

PR #17, opened 2023-05-24
feat: add user_agent to options

### LINKED ISSUE #16: Allow easy configuration of the user agent
For the API requests we should make it possible to set a custom user agent (and also encourage users to set the user agent for their apps).
This can be a setting on the client instance or class or the auth client.

im-adithya:
Fixes #16 
Changes the rest function to have user-agent as param, also supports passing user_agent header as usual

(If both are passed, the one passed as the header will be considered)

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -80,12 +80,12 @@ export async function request({
         ...(isPost
           ? { "Content-Type": "application/json; charset=utf-8" }
           : undefined),
+        ...authHeader,
+        ...headers,
         ...{
           "User-Agent": user_agent ?? "alby-js-api",
           "X-User-Agent": user_agent ?? "alby-js-api"
         },
-        ...authHeader,
-        ...headers,
       },
       method,
       body: isPost ? JSON.stringify(request_body) : undefined,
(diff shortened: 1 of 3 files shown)


bumi:
I think there is still a bug in Chrome that does not allow setting the user agent (https://bugs.chromium.org/p/chromium/issues/detail?id=571722)
because the SDK can be used in a browser environment let's additionally set a "X-User-Agent" header.

im-adithya:
Done!

src/request.ts
```diff
@@ -78,6 +80,9 @@ export async function request({
         ...(isPost
           ? { "Content-Type": "application/json; charset=utf-8" }
           : undefined),
+        ...(user_agent
+          ? { "User-Agent": user_agent, "X-User-Agent": user_agent }
+          : undefined),
```
rolznz, line 85: @im-adithya should we do the fallback here instead so it's impossible not to pass a user agent?

bumi: +1

im-adithya: Done!

PR #20, opened 2023-06-24
feat: add budget options to initNWC

rolznz:


src/webln/NostrWeblnProvider.ts
```diff
@@ -277,10 +278,26 @@ export class NostrWebLNProvider implements WebLNProvider {
     if (options?.returnTo) {
       url.searchParams.set('return_to', options.returnTo);
     }
+
+    if (options.budgetRenewal) {
+      url.searchParams.set("budget_renewal", options.budgetRenewal)
+    }
+    if (options.expiresAt) {
+      url.searchParams.set("expires_at", options.expiresAt.toISOString())
```
bumi, line 286: what do you think about using a timestamp instead of the string repesentation?

src/webln/NostrWeblnProvider.ts
```diff
@@ -277,10 +278,26 @@ export class NostrWebLNProvider implements WebLNProvider {
     if (options?.returnTo) {
       url.searchParams.set('return_to', options.returnTo);
     }
+
+    if (options.budgetRenewal) {
```
bumi, line 282: a few lines above we use `options?` should `options` be required?

bumi (approved):

src/webln/NostrWeblnProvider.ts
rolznz: @bumi I think it would be good to support both. Would that be milliseconds (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now) or seconds since the epoch?

nostr-wallet-connect only supports a string representation right now. I will create an issue there. I think this could be improved in a follow up PR?

src/webln/NostrWeblnProvider.ts
rolznz: Oops, I changed it back to optional. Thanks!

src/webln/NostrWeblnProvider.ts
bumi: both is just adding complexity but not really making it much better. It is easy to make a timestamp in any language and a timestamp also does not have the confusion of potentially different string representations with wrong timezones for example. 

seconds is the default typically afaik - and good enough.

rolznz: @bumi updated

PR #23, opened 2023-07-03
fix: refresh access token race condition

rolznz:
Fixes https://github.com/getAlby/alby-js-sdk/issues/22

View no-whitespace diff: https://github.com/getAlby/alby-js-sdk/pull/23/files?diff=unified&w=1


rolznz:
@bumi confirmed google also uses a similar solution to prevent duplicate requests to get a new access token

bumi:
@rolznz can you share some links?

rolznz:
Google client: https://github1s.com/googleapis/google-auth-library-nodejs/blob/main/src/auth/oauth2client.ts#L667-L669

PR #24, opened 2023-07-03
fix: make NWC provider correctly implement NIP-07 interface

rolznz:


rolznz:
As @bumi mentioned NWC uses its own keypair which probably shouldn't be used for anything else?

I am not sure this really makes sense any more?

bumi:
as we have a related object that implements exactly those methods I still think it makes sense to implement the same API. 
for whatever people want to use it. 

this is just a breaking change, so we should release it as a major version

src/webln/NostrWeblnProvider.ts
```diff
@@ -115,11 +120,16 @@ export class NostrWebLNProvider implements WebLNProvider {
     return getPublicKey(this.secret);
   }
 
-  signEvent(event: UnsignedEvent) {
+  getPublicKey(): Promise<string> {
+    return Promise.resolve(this.publicKey);
+  }
+
+  signEvent(event: UnsignedEvent): Promise<Event> {
```
bumi, line 127: as you change the signature of this method you also need to adjust the methods where this is used.

rolznz: There are no usages outside of this class (apart from being a breaking change for dependencies)

bumi: https://github.com/getAlby/alby-js-sdk/blob/master/src/webln/NostrWeblnProvider.ts#L187

rolznz: @bumi sorry about that 😱

PR #26, opened 2023-07-06
fix: set message on alby response error and export it

rolznz:
fixes https://github.com/getAlby/alby-js-sdk/issues/25

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,3 +1,4 @@
+import { AlbyResponseError } from "./AlbyResponseError";
 
 export type SuccessStatus = 200 | 201;
 export type ResponseType = "application/json";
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -223,3 +223,23 @@ export type GetAccountInformationResponse = {
...
+  statusText: string;
+  headers: Record<string, any>;
+  error: any; // todo: typeable?
+  constructor(
+    status: number,
+    statusText: string,
+    headers: Headers,
+    error: any
+  ) {
+    super([status.toString(), statusText, error.message ? error.message : JSON.stringify(error)].filter(text => text).join(" "));
```
bumi, line 239: this line is a bit fancy. I think it would be easier to read when it's split up in multiple lines

PR #33, opened 2023-07-22
feat: swaps

reneaaron:
Adds methods to get swap info & create new swaps.

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -58,14 +58,13 @@ export async function request({
     url.toString(),
     {
       headers: {
-        ...(isPost
-          ? { "Content-Type": "application/json; charset=utf-8" }
-          : undefined),
+        ...
+        { "Content-Type": "application/json; charset=utf-8" },
         ...authHeader,
         ...headers,
         ...{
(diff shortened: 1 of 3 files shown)


src/request.ts
```diff
@@ -58,14 +58,13 @@ export async function request({
     url.toString(),
     {
       headers: {
-        ...(isPost
-          ? { "Content-Type": "application/json; charset=utf-8" }
-          : undefined),
+        ...
+        { "Content-Type": "application/json; charset=utf-8" },
```
rolznz, line 62: is it ok to always set the content type even if we don't post a body? I guess it doesn't matter right?

rolznz (approved):

rolznz:
Looks great to me!

src/request.ts
reneaaron: I can't remember why I added it :thinking: 

According to https://stackoverflow.com/questions/5661596/do-i-need-a-content-type-header-for-http-get-requests it should only be added to POST / PUT requests. 

I'll try to revert.

PR #34, opened 2023-07-24
feat: setup outDir Directory

pavanjoshi914:
Cannot write file '/home/pavan/Desktop/codebases/getAlby/alby-js-sdk-master/alby-js-sdk/src/window.js' because it would overwrite input file.

image

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -12,7 +12,8 @@
     "target": "es2020",
     "rootDir": "./src",
     "sourceMap": true,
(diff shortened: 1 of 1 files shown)


bumi:
how does that relate to `microbundle` which also uses `dist` I guess?

pavanjoshi914:
okay. microbundle handles this automatically. i made this quickly just to resolve type script error. but i reforked the repo and error went away. closing this for now

PR #36, opened 2023-07-26
chore: rename alby js sdk package

rolznz:
Follow up steps:
- [ ] Deprecate old npm package
- [ ] Announce package changes in Alby blog

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@getalby/sdk",
-  "version": "1.0.0",
+  "version": "2.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,12 +1,12 @@
 {
-  "name": "alby-js-sdk",
-  "version": "2.1.0",
+  "name": "@getalby/sdk",
+  "version": "1.0.0",
```
bumi, line 3: can we keep the version?

bumi:
the user agent could also be updated

PR #37, opened 2023-07-26
feat: make alby oauth url customizable

pavanjoshi914:


src/types.ts
```diff
@@ -28,12 +28,14 @@ export type GenerateAuthUrlOptions =
   | {
       code_challenge_method?: string;
       code_challenge?: string;
+      testnet: boolean;
```
rolznz, line 31: I think we should pass an `authorizeUrl?: string` here for better flexibility, and make it optional.

Also can you update the type so it is `{authorizeUrl?: string} &( A | B | C)` so you don't need to duplicate the declaration

src/OAuth2User.ts
```diff
@@ -3,7 +3,7 @@ import { buildQueryString, basicAuthHeader } from "./utils";
 import { OAuthClient, AuthHeader, GetTokenResponse, Token, GenerateAuthUrlOptions } from "./types";
 import { RequestOptions, rest } from "./request";
 
-const AUTHORIZE_URL = "https://getalby.com/oauth";
+let AUTHORIZE_URL = "https://getalby.com/oauth";
```
rolznz, line 6: could you please keep this as a constant?

src/OAuth2User.ts
```diff
@@ -162,6 +161,10 @@ export class OAuth2User implements OAuthClient {
       code_challenge_method = "plain";
     }
     const code_challenge = this.code_challenge;
+    if(options.testnet){
+      AUTHORIZE_URL="https://app.regtest.getalby.com/oauth"
+
+    }
     const url = new URL(AUTHORIZE_URL);
```
rolznz, line 168: I think `options.authorizeUrl || AUTHORIZE_URL` is better, so we can keep the existing one as a constant and default value

src/types.ts
pavanjoshi914: in that case 
1. this also requires validation. if client is using right url or not

pavanjoshi914: user has only two optoins. default mainnet. if they wish to setup testnet they just set a boolean. isn't this more convenient?

rolznz: We could also point it at a local running instance for example. @bumi do you have any preference here?

pavanjoshi914: resolved

src/OAuth2User.ts
pavanjoshi914: resolved

src/OAuth2User.ts
pavanjoshi914: resolved

src/types.ts
bumi: +1 for setting the URL

pavanjoshi914: made suggested changes

rolznz (approved):
utACK

pavanjoshi914:
do we need a validation of url here?

bumi:
> do we need a validation of url here?

no it's for developers not end-users. So they are smart and can deal with errors that would happen.
and validation just requires code that needs to be maintained, makes the package more complicated (and any line of code will break at some point, so better avoid it)

pavanjoshi914:
i have tested it with the yarn link things work well!

bumi:
great. can you add a note of the option to the readme?

README.md
```diff
@@ -225,6 +225,7 @@ const authClient = new auth.OAuth2User({
 
 const authUrl = authClient.generateAuthURL({
   code_challenge_method: "S256",
+  authorizeUrl: "https://getalby.com/oauth" // endpoint for authorization (replace with the appropriate URL based on the environment)
```
rolznz, line 228: if it's put in the example like this, users will think they need to pass it. Should it be commented out here?

bumi:
and actually we need two options: the authorize URL and the API URL?

rolznz:
> and actually we need two options: the authorize URL and the API URL?

@bumi the API url is already configurable (it is called base url)

README.md
pavanjoshi914: resolved

PR #39, opened 2023-07-27
fix: build query string util function double &

rolznz:


bumi:
shouldn't here `URLSearchParams` actually work?

(new URLSearchParams({foo:1, bar: 3})).toString()

rolznz:
@bumi it doesn't omit undefined values:
```
new URLSearchParams({a: 1, b:2, c: undefined}).toString()
'a=1&b=2&c=undefined'
```

PR #40, opened 2023-07-27
fix: also check if client_secret is provided in request access token method

rolznz:


bumi:
I don't think the secret is always required. In the PKCE mode it is not.

rolznz:
@bumi thanks, could you please look again?

PR #42, opened 2023-08-03
token changed callback

### LINKED ISSUE #31: Add token changed callback
Currently it is difficult to figure out when the refresh/access token pair changes (since it happens internally on request), which needs to be stored by the client app.

I found https://github.com/googleapis/google-api-nodejs-client seems to manage refreshing the token automatically for the client and exposes an .on('tokens') callback so apps can subscribe to token changes and store the credentials themselves (for next time the app is initialized).

Can we do the same here and update the documentation in the README so it is obvious how to manage tokens?
- Tokens are updated automatically
- Use this method to subscribe to changes and store the token yourself

pavanjoshi914:
Fixes #31

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ The Alby JS SDK allows you to easily integrate Nostr Wallet Connect into any Jav
 
 The `NostrWebLNProvider` exposes the [WebLN](webln.guide/) sendPayment interface to execute lightning payments through Nostr Wallet Connect.
 
-(note: in the future more WebLN functions will be added to Nostr Wallet Connect)
+(Note: in the future more WebLN functions will be added to Nostr Wallet Connect)
 
 ### NostrWebLNProvider (aliased as NWC) Options
 
@@ -256,8 +256,13 @@ const client = new Client(authClient);
 const result = await client.createInvoice({amount: 1000});
 ```
 
-#### Handling refresh token 
-Access tokens do expire. If an access token is about to expire, this library will automatically use a refresh token to retrieve a fresh one. Utilising the tokens event is a simple approach to guarantee that you always save the most recent tokens:
+#### Handling refresh token
+
+Access tokens do expire. If an access token is about to expire, this library will automatically use a refresh token to retrieve a fresh one. Utilising the *tokens* event is a simple approach to guarantee that you always save the most recent tokens.
+
+If token refresh fails, you can restart the OAuth Authentication flow or log the error by listening for the *tokenRefreshFailed* event.
+
+(Note: To prevent losing access to the user's token, only initialize one instance of the client per token pair at a time)
 
 ```js
 const token = loadTokenForUser(); // {access_token: string, refresh_token: string, expires_at: number}
@@ -270,11 +275,14 @@ const authClient = new auth.OAuth2User({
 
 // listen to the tokens event
 authClient.on('tokens', (tokens) => {
-  if (tokens.refresh_token) {
     // store the tokens in database
-    console.log(tokens.refresh_token);
-  }
-  console.log(tokens.access_token);
+    console.log(tokens);
+});
+
+// Listen to the tokenRefreshFailed event
+authClient.on('tokenRefreshFailed', (error) => {
+    // Handle the token refresh failure, for example, log the error or launch OAuth authentication flow
+    console.error("Token refresh failed:", error.message);
 });
 ```
 
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ const result = await client.createInvoice({amount: 1000});
 
 #### Handling refresh token
 
-Access tokens do expire. If an access token is about to expire, this library will automatically use a refresh token to retrieve a fresh one. Utilising the *tokenRefresh* event is a simple approach to guarantee that you always save the most recent tokens.
+Access tokens do expire. If an access token is about to expire, this library will automatically use a refresh token to retrieve a fresh one. Utilising the *tokenRefreshed* event is a simple approach to guarantee that you always save the most recent tokens.
 
 If token refresh fails, you can restart the OAuth Authentication flow or log the error by listening for the *tokenRefreshFailed* event.
 
@@ -273,8 +273,8 @@ const authClient = new auth.OAuth2User({
   token: token
 });
 
-// listen to the tokenRefresh event
-authClient.on('tokenRefresh', (tokens) => {
+// listen to the tokenRefreshed event
+authClient.on('tokenRefreshed', (tokens) => {
     // store the tokens in database
     console.log(tokens);
 });
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ const result = await client.createInvoice({amount: 1000});
 
 #### Handling refresh token
 
-Access tokens do expire. If an access token is about to expire, this library will automatically use a refresh token to retrieve a fresh one. Utilising the *tokens* event is a simple approach to guarantee that you always save the most recent tokens.
+Access tokens do expire. If an access token is about to expire, this library will automatically use a refresh token to retrieve a fresh one. Utilising the *tokenRefresh* event is a simple approach to guarantee that you always save the most recent tokens.
 
 If token refresh fails, you can restart the OAuth Authentication flow or log the error by listening for the *tokenRefreshFailed* event.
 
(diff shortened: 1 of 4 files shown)


src/OAuth2User.ts
```diff
@@ -39,13 +40,17 @@ export class OAuth2User implements OAuthClient {
...
+  private _tokenEvents = new EventEmitter();
   constructor(options: OAuth2UserOptions) {
     const { token, ...defaultOptions } = options;
     this.options = {client_secret: '', ...defaultOptions};
     this.token = token;
     this._refreshAccessTokenPromise = null;
   }
-
+  
+  on(eventName: string, listener: (tokens: Token) => void) {
```
rolznz, line 51: can you type the eventName so it only accepts the ones we support - `eventName: "tokens"`

rolznz, line 51: Can you please type the return type. I don't think we should return this._tokenEvents; 

Could it be `: Promise<void>` and then change the below line to:
`await this._tokenEvents.on(eventName, listener);`

rolznz:
@pavanjoshi914 can you please update the README with a section on how to subscribe to token changes?

src/OAuth2User.ts
pavanjoshi914: resolved, but why we need await. after doing a promise<void>

rolznz: Right now the function won't wait for the .on async function to complete. Can you change it to one of these:
```
async on(eventName: EventName, listener: (tokens: Token) => void): Promise<void> {
    await this._tokenEvents.on(eventName, listener);
  }```
  
or (I think this one is better):
```
on(eventName: EventName, listener: (tokens: Token) => void): Promise<void> {
    return this._tokenEvents.on(eventName, listener);
  }
```

File: README.md [hunk 2]
```diff
@@ -256,6 +256,28 @@ const client = new Client(authClient);
 const result = await client.createInvoice({amount: 1000});
 ```
 
+#### Handling refresh token 
+Access tokens do expire. If an access token is about to expire, this library will automatically use a refresh token to retrieve a fresh one. Utilising the tokens event is a simple approach to guarantee that you always save the most recent tokens:
```
rolznz, line 260: Could we add a note here: make sure to only initialize one instance of the client per token pair at a time, otherwise you could lose access to the user's token.

@bumi for backends that are scaled to more than one instance, this could still be an issue, right?

README.md
```diff
@@ -256,6 +256,28 @@ const client = new Client(authClient);
...
+const authClient = new auth.OAuth2User({
+  client_id: process.env.CLIENT_ID,
+  callback: "http://localhost:8080/callback",
+  scopes: ["invoices:read", "account:read", "balance:read", "invoices:create", "invoices:read", "payments:send"],
+  token: token
+});
+
+// listen to the tokens event
+authClient.on('tokens', (tokens) => {
+  if (tokens.refresh_token) {
```
rolznz, line 273: why do we need a check here? the event should always provide a valid token.

Why log twice?

We should add another event for failed token refresh, right?

src/OAuth2User.ts
pavanjoshi914: but we need to return promise on  function inside on is complete. so we do need async and await.

README.md
pavanjoshi914: resolved

README.md
pavanjoshi914: can we do it in the same tokens event, that refreshAccessToken do. ```Promise<{ token: Token }>``` return this, if token refresh is done. else return null. the user stores the tokens if not. it handles the error by relaunching oauth flow? 


if token refresh is failed we will be still returning expired token that already has in ```this.token```. handling error and getting toking, two different methods related to same purpose,  wouldn't  be little inconsistent?

src/OAuth2User.ts
```diff
@@ -84,8 +98,11 @@ export class OAuth2User implements OAuthClient {
         const token = processTokenResponse(data);
         this.token = token;
         resolve({token});
+        this._tokenEvents.emit("tokens", this.token);
```
bumi, line 101: should we also add a verb here? (like in tokenRefresh, accountChange, ...)

pavanjoshi914: resolved

src/OAuth2User.ts
```diff
@@ -23,6 +24,11 @@ export interface OAuth2UserOptions {
   token?: Token;
 }
 
+export type TokenListener = (tokens: Token) => void;
```
rolznz, line 27: Should this be `TokenRefreshListener`?

src/OAuth2User.ts
```diff
@@ -23,6 +24,11 @@ export interface OAuth2UserOptions {
...
+export type TokenRefreshFailedListener = (error: Error) => void;
+export type EventName= "tokenRefresh" | "tokenRefreshFailed";
```
rolznz, line 29: Can you rename to `TokenEventName`?

rolznz: Or will we have other events?

src/OAuth2User.ts
```diff
@@ -39,13 +45,21 @@ export class OAuth2User implements OAuthClient {
...
     const { token, ...defaultOptions } = options;
     this.options = {client_secret: '', ...defaultOptions};
     this.token = token;
     this._refreshAccessTokenPromise = null;
   }
 
+    /**
+   * Subscribe to the events
+   */
+  async on(eventName: EventName, listener: Listener): Promise<void> {
```
rolznz, line 59: Is there an `off` method? (not sure it is needed?)

pavanjoshi914: do we nee that?. client can just remove the code for ```.on```  if they don't want to listen to the token changes and process it

src/OAuth2User.ts
pavanjoshi914: > good idea, this then should also should be used by the webln provider and exposed there [master/src/webln/OauthWeblnProvider.ts#L29-L38](https://github.com/getAlby/alby-js-sdk/blob/master/src/webln/OauthWeblnProvider.ts?rgh-link-date=2023-07-20T09%3A10%3A57Z#L29-L38)

how the webln provider will use it? I think keeping it general will be good

src/OAuth2User.ts
rolznz: How do they remove the code of .on?

pavanjoshi914: I mean they just don't listen using on method. cause the callback is fixed here. which passes the token or an error

pavanjoshi914: they will not call the on method if they don't want to listen

pavanjoshi914: cc @bumi wdyt? or shall we implement it? just in case for future reference?

src/OAuth2User.ts
pavanjoshi914: resolved

src/OAuth2User.ts
pavanjoshi914: resolved. we don't need this currently

rolznz (approved):

rolznz:
@pavanjoshi914 can you test these changes with the Alby extension using yarn link? using these callbacks should be much cleaner than the current solution.

README.md
```diff
@@ -256,6 +256,36 @@ const client = new Client(authClient);
...
+const token = loadTokenForUser(); // {access_token: string, refresh_token: string, expires_at: number}
+const authClient = new auth.OAuth2User({
+  client_id: process.env.CLIENT_ID,
+  callback: "http://localhost:8080/callback",
+  scopes: ["invoices:read", "account:read", "balance:read", "invoices:create", "invoices:read", "payments:send"],
+  token: token
+});
+
+// listen to the tokenRefresh event
+authClient.on('tokenRefresh', (tokens) => {
```
rolznz, line 277: Minor: `tokenRefreshed` would be better since it has already been refreshed, and then it is consistent with `tokenRefreshFailed`.

src/OAuth2User.ts
```diff
@@ -39,13 +45,21 @@ export class OAuth2User implements OAuthClient {
   code_verifier?: string;
   code_challenge?: string;
   private _refreshAccessTokenPromise: Promise<{token: Token}> | null;
+  private _tokenEvents = new EventEmitter();
```
rolznz, line 48: for consistency should this be initialized in the constructor?

pavanjoshi914:
@rolznz @bumi  i tested it i can receive token from the method. are we good to merge this?

bumi:
good for me. @rolznz feel free to merge it.

rolznz (approved):

PR #60, opened 2023-08-10
feat: add linters

### LINKED ISSUE #43: Add prettier and linting
It would be nice to add prettier and pre-commit linting (using husky). This way later commits do not keep changing the formatting. 

Maybe the extension can be looked at as an example.

pavanjoshi914:
add prettier 
add husky to create pre commit hooks
add lint-staged pre commit hook for prettier
move jsx file to examples folder
add eslint
add eslint typescript parser
add config plugin to run eslint and prettier serially

Fixes #43 
signed-off-by: pavan joshi <pavanj914@gmail.com>


pavanjoshi914:
@rolznz

lint-staged.config.js
```diff
@@ -0,0 +1,13 @@
+export default {
```
rolznz, line 1: I personally do not like lint-staged, it requires new commands run here instead of the ones in package.json and for me it seems to not always work correctly. I know this is in the extension I would prefer we just lint all outstanding changes. Do you have any thoughts here @reneaaron @bumi ?

rolznz:
@pavanjoshi914 the changes look good to me. But surely there will be more file changes after we actually run those commands (for example, running prettier)? did the commit checks pass for you?

rolznz:
@pavanjoshi914 please also add a new PR off of this branch and run the linting so we can see the changes

rolznz (approved):

lint-staged.config.js
rolznz: decision is to leave as is. Lint-staged runs faster and we can run full lint on CI

PR #65, opened 2023-08-14
doc: make loadNWCUrl example more clear

rolznz:
Two users so far have asked about this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ The `NostrWebLNProvider` exposes the [WebLN](webln.guide/) sendPayment interface
 * `authorizationUrl`: URL to the NWC interface for the user to and the app connection
 
 ### Quick start example
-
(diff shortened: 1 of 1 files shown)


pavanjoshi914:
I kinda always ignore notes thinking that should be optional :smiley:  if something is mandatory information I prefer to mention it directly or in a separate section so that the user can have its attention on it! but really depends

like here we are asking user to create nwc url, and then again a second link and then again implement function by yourself

rolznz:
@pavanjoshi914 I bolded the note, but it is not perfect. Do you have a better idea?

bumi:
+1 in docs we can expect people to read docs. If a developer ignores the docs and does not read it it is their fault.
we can not build their apps, they have to do semething.

we can link to the examples  - and also add some more examples.

rolznz:
Ok, I will close this for now. We can provide support for devs who get stuck.

PR #73, opened 2023-08-21
feat: add lookup invoice to nwc webln provider

rolznz:
this is not actually a webln method. Not sure what to do here.

### DIFF
--- a/src/webln/NostrWeblnProvider.ts
+++ b/src/webln/NostrWeblnProvider.ts
@@ -20,7 +20,6 @@ import {
   SignMessageResponse,
   WebLNNode,
   WebLNProvider,
-  WebLNMethod,
   WebLNRequestMethod,
   LookupInvoiceArgs,
   LookupInvoiceResponse,
@@ -47,6 +46,13 @@ interface NostrWebLNOptions {
 type Nip07Provider = {
   getPublicKey(): Promise<string>;
   signEvent(event: UnsignedEvent): Promise<Event>;
+};
+
+const nip47ToWeblnRequestMap = {
+  get_balance: "getBalance",
+  make_invoice: "makeInvoice",
+  pay_invoice: "sendPayment",
+  lookup_invoice: "lookupInvoice",
 };
 
 export class NostrWebLNProvider implements WebLNProvider, Nip07Provider {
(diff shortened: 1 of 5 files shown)


src/webln/NostrWeblnProvider.ts
```diff
@@ -283,6 +294,19 @@ export class NostrWebLNProvider implements WebLNProvider, Nip07Provider {
       (result) => ({ paymentRequest: result.invoice }),
     );
   }
+
+  lookupInvoice(args: LookupInvoiceArgs) {
+    this.checkConnected();
+
+    return this.executeNip47Request<LookupInvoiceResponse>(
+      "lookup_invoice",
+      "lookupinvoice",
```
bumi, line 303: do we actually need to pass the two here? or can this be some mapping? 
this is only needed for the events, isn't it?

src/webln/NostrWeblnProvider.ts
```diff
@@ -214,7 +224,7 @@ export class NostrWebLNProvider implements WebLNProvider, Nip07Provider {
   // TODO: use NIP-47 get_info call
   async getInfo(): Promise<GetInfoResponse> {
     return {
-      methods: ["getInfo", "sendPayment", "addinvoice", "getBalance"],
+      methods: ["getInfo", "sendPayment", "addinvoice", "getBalance", "lookupinvoice"],
```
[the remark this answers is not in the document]
rolznz: @bumi thanks for confirming, I got a bit confused here.

The ones I put are LND methods, but then we also have the official WebLN methods.

src/webln/NostrWeblnProvider.ts
rolznz: @bumi the webln request method is needed for the notification function. WebLN and NWC event names are inconsistent, I think we need to provide both.

Unless we can guarantee the only change is in casing, then we could have a mapping function from camelCase to lower_case_with_underscores

src/webln/NostrWeblnProvider.ts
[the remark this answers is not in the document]
rolznz: @bumi I have updated webln-types and fixed those inconsistencies. Thanks!

src/webln/NostrWeblnProvider.ts
bumi: but can't we do a mapping from 

```
a = { 
  "lookup_invoice" : "lookupinvoice" 
}
a["lookup_invoice"]
```

that then can be checked?

PR #82, opened 2023-08-29
feat: convert balance from msat

rolznz:
Should be merged as soon as new version of Alby NWC is deployed.

### DIFF
--- a/src/webln/NostrWeblnProvider.ts
+++ b/src/webln/NostrWeblnProvider.ts
@@ -232,11 +232,21 @@ export class NostrWebLNProvider implements WebLNProvider, Nip07Provider {
   getBalance() {
     this.checkConnected();
 
-    return this.executeNip47Request<GetBalanceResponse, { balance: number }>(
+    return this.executeNip47Request<
+      GetBalanceResponse,
+      { balance: number; max_amount?: number; budget_renewal?: string }
+    >(
       "get_balance",
       undefined,
       (result) => result.balance !== undefined,
-      (result) => result,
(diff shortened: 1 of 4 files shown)


src/webln/NostrWeblnProvider.ts
```diff
@@ -232,11 +232,22 @@ export class NostrWebLNProvider implements WebLNProvider, Nip07Provider {
...
+      { balance: number; max_amount?: number; budget_renewal?: string }
+    >(
       "get_balance",
       undefined,
       (result) => result.balance !== undefined,
-      (result) => result,
+      (result) => ({
+        // NWC uses msats - convert to sats for webln
+        balance: Math.floor(result.balance / 1000),
+        max_amount:
```
rolznz, line 245: does it make sense to return a separate field here for WebLN? or should we just return the balance as the max_amount if it exists?

Edit: I think this is still wrong. Example response:
```{ balance: 10101000, max_amount: 100000, budget_renewal: 'daily' }```

This does not show how much of the budget was used.

bumi: Afaik there is no spec for "budgets" and those can vary a lot. I don't think this is possible.

imo it should be up to the wallet to decide on how it wants to define "balance" and we should only use "balance" here.

rolznz: updated. I noticed there was a currency field in the spec at https://webln.guide/ so I hardcoded it as sats for now. CC @reneaaron

PR #101, opened 2023-10-12
feat: add invoices method

pavanjoshi914:
add invoices method to fetch all the transactions

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -10,6 +10,7 @@ import {
   CreateWebhookEndpointResponse,
   GetAccountBalanceResponse,
   GetAccountInformationResponse,
+  GetInvoicesRequestParams,
   Invoice,
   InvoiceRequestParams,
   KeysendRequestParams,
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -113,6 +113,20 @@ export class Client {
     });
   }
 
+  invoices(
+    params: {},
```
bumi, line 117: as params we can pass in filters. do those need some type defenition?

pavanjoshi914: what kind of filters we support? passing limit of invoices that can be retrieved as params ?

reneaaron: This is the list of supported params: 

https://guides.getalby.com/alby-wallet-api/reference/api-reference/invoices

image

pavanjoshi914: image

cool! but we didn't added support for params in api docs. we shall update that as well if this endpoint supports params

bumi: we should update those, yes. can you do it? copy it from the other endpoint.

reneaaron: I've updated gitbook: 

https://guides.getalby.com/alby-wallet-api/reference/api-reference/invoices#get-all-invoice-history

reneaaron: I've also added some typed params for the invoices endpoint now.

rolznz (approved):
utACK

pavanjoshi914:
thanks @reneaaron for the params stuff! merging this

Issue #120, opened 2023-11-02
requestAccessToken does not emit tokenRefreshed function


rolznz:
This means the user of the sdk needs to handle the access tokens in two different ways (initial token vs refreshed). I think it should fire the same event.

Currently an issue in this PR: https://github.com/getAlby/lightning-browser-extension/pull/2830

rolznz:
Closing, see https://github.com/getAlby/js-sdk/pull/121

PR #121, opened 2023-11-03
feat: deliver token when generated for the first time

pavanjoshi914:
https://github.com/getAlby/js-sdk/issues/120


rolznz (approved):
utACK

rolznz:
Chatted with @pavanjoshi914 and @bumi and it seems more straightforward to keep as is, rather than relying on this side effect (especially when using the first token)

PR #122, opened 2023-11-03
feat: use user agent for requests from auth object

pavanjoshi914:


src/request.ts
```diff
@@ -52,6 +52,11 @@ export async function request({
   const authHeader = auth
     ? await auth.getAuthHeader(url.href, method)
     : undefined;
+
+  let userAgent;
+  if(auth?.getUserAgent)
```
rolznz, line 57: this is not needed

`this.defaultRequestOptions` from the Client class is passed to `rest` which is passed to this `request` method.

You don't need to add a function `getUserAgent`.

Can you please undo the change, test this locally and make sure the user agent is being passed in every request, and if not, why?

pavanjoshi914:
closing it as we now decided to use user-agent passed in default request options object and not auth object

Issue #127, opened 2023-11-11
Add support for decoding invoices


rolznz:
The Alby JS SDK has a public endpoint that can be used: https://guides.getalby.com/developer-guide/v/alby-wallet-api/reference/api-reference/invoices#decode-an-invoice

rolznz:
For now try light-bolt11-decoder (review added dependencies)

PR #131, opened 2023-11-14
feat: add decode invoice function

rolznz:
fixes https://github.com/getAlby/js-sdk/issues/127

adds only one dependency

- [ ] also add to [js-lightning-tools](https://github.com/getAlby/js-lightning-tools/issues/44)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,3 +2,4 @@ export * as auth from "./auth";
 export * as types from "./types";
 export * as webln from "./webln";
 export { Client } from "./client";
+export * from "./decodeInvoice";
--- a/src/light-bolt11-decoder.d.ts
+++ b/src/light-bolt11-decoder.d.ts
@@ -1,0 +1,90 @@
+// TODO: submit PR to https://github.com/nbd-wtf/light-bolt11-decoder
+declare module "light-bolt11-decoder" {
+  type NetworkSection = {
+    name: string;
+    letters: string;
+    value?: {
+      bech32: string;
+      pubKeyHash: number;
+      scriptHash: number;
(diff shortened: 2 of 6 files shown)


src/light-bolt11-decoder.d.ts
```diff
@@ -0,0 +1,90 @@
+// TODO: submit PR to https://github.com/nbd-wtf/light-bolt11-decoder
```
rolznz, line 1:
```suggestion
// TODO: remove when https://github.com/nbd-wtf/light-bolt11-decoder/pull/4 is merged
```

src/index.ts
```diff
@@ -2,3 +2,4 @@ export * as auth from "./auth";
 export * as types from "./types";
 export * as webln from "./webln";
 export { Client } from "./client";
+export * from "./decodeInvoice";
```
bumi, line 5: hmmm, I don't think this is a nice API. 
Just exposing the global function because that people don't have to import it from the light-bolt11-decoder. 

doesn't this make more sense to have in the `Invoice` https://github.com/getAlby/js-lightning-tools/blob/master/src/invoice.ts

So I can initialize an invoice and then have all the data available there?

rolznz: @bumi sounds good to me. Would it make sense for this sdk to depend on lightning-tools? or should people install both packages as they are doing now? I think it would be nice to have some support for decoding invoices in this library?

rolznz: Update: we could add a small link to the README

# Decoding Invoices
See [lightning-tools](https://github.com/getAlby/js-lightning-tools)

PR #132, opened 2023-11-15
feat: add preimage

reneaaron:


rolznz (approved):
utACK

src/types.ts
```diff
@@ -206,6 +206,7 @@ export type Invoice = {
   payer_pubkey?: string;
   payment_hash: string;
   payment_request: string;
+  preimage: string;
```
bumi, line 209: it might also be null

rolznz: what case is it null right now? these are only settled transactions, right?

But in the future they might not be all settled. Fix here: https://github.com/getAlby/js-sdk/pull/136

bumi: true the API right now only returns settled invoices.

PR #136, opened 2023-11-17
fix: make preimage optional

rolznz:
it will not exist for unsettled payments (currently not returned from the API)


reneaaron (approved):
Oh no, obviously :sweat_smile: Thanks for fixing it! 

utACK

Issue #56, opened 2023-08-09
Update nostr tools


rolznz:
There was a breaking change so we have currently fixed to the version before the change

PR #137, opened 2023-11-17
chore: update nostr tools

rolznz:
Fixes https://github.com/getAlby/js-sdk/issues/56


bumi:
we had issues with the update before, but this works?

rolznz:
@bumi I tested NWC, it works.

I don't think we had issues, we just didn't update it due to nostr-tools having a breaking change without a correct version bump which unexpectedly broke for people who installed alby-js-sdk at that time, right?

PR #139, opened 2023-11-17
chore: rename sendToAlbyAccount to sendBoostagramToAlbyAccount

rolznz:
**BREAKING CHANGE**

This is a keysend request so I think this naming makes more sense

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ Please have a look a the Alby OAuth2 Wallet API:
 - keysend
 - sendPayment
 - sendBoostagram
-- sendToAlbyAccount
+- sendBoostagramToAlbyAccount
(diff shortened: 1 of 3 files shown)


bumi:
I think we should deprecate it first. 
meaning that the old method logs a deprecation warning that it will be removed soon.

rolznz:
@bumi good point, fixed. Thanks!

bumi:
Did you do a `console.log` ? or is that bad - because people will not read that code.

bumi:
do you know the JS best practices for this? what do others do?

rolznz:
I think we did the main things. I will update the comment to say it will be removed in version 3.0.0

https://stackoverflow.com/a/19525656

PR #140, opened 2023-11-17
feat: decode invoice

rolznz:
- [ ] in the example a scope is needed to create a client even though the scope is not actually needed and does not make sense. How would we do this in the future? add a new scope?
- [ ] unlike the lightning tools package, there is no mapping of the response type. Should we do basic mapping here again to e.g. expose Javascript dates for the invoice created date and expiry?

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,7 @@ Please have a look a the Alby OAuth2 Wallet API:
 - outgoingInvoices
 - getInvoice
 - createInvoice
+- decodeInvoice
 - keysend
 - sendPayment
 - sendBoostagram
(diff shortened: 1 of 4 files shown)


bumi:
> in the example a scope is needed to create a client even though the scope is not actually needed and does not make sense. How would we do this in the future? add a new scope?

I think this is fine. If you use the SDK and also have an access token then you have some scope. 
Otherwise you can just do `fetch("https://api.getalby.com/.../decode/[PR]")` and you don't need this SDK. 


> unlike the lightning tools package, there is no mapping of the response type. Should we do basic mapping here again to e.g. expose Javascript dates for the invoice created date and expiry?

either everywhere or nowhere imo. I guess we don't do it at the other endpoints also.

bumi (approved):
utACK

PR #152, opened 2023-12-04
feat: add support for new nwc extensions

im-adithya:
⚠️ Currently fails since it requires a change in webln types for the `GetInfoResponse`

## Description

Adds methods to support the upcoming NWC extensions: `get_info`, `pay_keysend`, `list_invoices`, `list_payments`


rolznz:
@im-adithya regarding this:
> Currently fails since it requires a change in webln types for the GetInfoResponse

I think this isn't an issue - we can have a NWCGetInfoResponse and map it to the original WebLN GetInfoResponse.

rolznz:
this was split up and merged individually

PR #172, opened 2023-12-21
fix: connect on every nwc request to ensure reconnect happens

rolznz:


src/webln/NostrWeblnProvider.ts
```diff
@@ -340,8 +337,8 @@ export class NostrWebLNProvider implements WebLNProvider, Nip07Provider {
     );
   }
 
-  keysend(args: KeysendArgs) {
-    this.checkConnected();
+  async keysend(args: KeysendArgs) {
+    await this.checkConnected();
 
     return this.executeNip47Request<SendPaymentResponse, Nip47PayResponse>(
```
bumi, line 343: do those need to be awaited now? 

is it good to mix async/await with returned promisses?

bumi:
This now changes the API - which might be OK, but now calling enable is no longer required.

shouldn't we still require that? and check if `enable` was called and if so we do call the connect function

src/webln/NostrWeblnProvider.ts
rolznz: It's just the same as doing
```
const result = await ...;
return result;
```

bumi: thx.

bumi (approved):

PR #190, opened 2024-01-22
feat: add sendMultiPayment function and example

rolznz:
~~This adds a dependency on lightning-tools to decode invoices, since the current spec returns payment hashes by default.
Or we could remove the dependency and generate ids and pass them to do the mapping, maybe that's better.~~ UPDATE: using `id` field of nwc to map instead of payment hash

The idea is that you can pay multiple invoices in a single request, but you need to know which ones succeeded and which ones failed. I was thinking an object like:

```
{
  payments: [
    {
      paymentRequest: 'lnbc...',
      preimage: '7c2c....'
    },
    ...,
  ],
  errors: [{
     paymentRequest: 'lnbc...',
     message: "Insufficient balance".
     code: 1234
  }]
}
```

Note: the error handling isn't done yet

### DIFF
--- a/src/webln/NostrWeblnProvider.ts
+++ b/src/webln/NostrWeblnProvider.ts
@@ -25,7 +25,6 @@ import {
 } from "@webbtc/webln-types";
 import { GetInfoResponse } from "@webbtc/webln-types";
 import { NWCAuthorizationUrlOptions } from "../types";
-import { Invoice } from "@getalby/lightning-tools";
 
 const NWCs: Record<string, NostrWebLNOptions> = {
   alby: {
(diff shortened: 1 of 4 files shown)


src/webln/NostrWeblnProvider.ts
```diff
@@ -644,6 +689,139 @@ export class NostrWebLNProvider implements WebLNProvider, Nip07Provider {
...
+                throw new Error(
+                  "Response from NWC failed validation: " +
+                    JSON.stringify(response.result),
+                );
+              }
+              const dTag = event.tags.find((tag) => tag[0] === "d")?.[1];
+              if (dTag === undefined) {
+                throw new Error("No d tag found in response event");
+              }
+              // console.info("dTag", dTag);
```
im-adithya, line 776: Should remove this

src/webln/NostrWeblnProvider.ts
```diff
@@ -644,6 +689,139 @@ export class NostrWebLNProvider implements WebLNProvider, Nip07Provider {
...
+                this.notify(weblnMethod, response.result);
+              }
+            } catch (error) {
+              console.error(error);
+              clearTimeout(replyTimeoutCheck);
+              sub.unsub();
+              reject({
+                error: (error as Error).message,
+                code: "INTERNAL",
+              });
```
im-adithya, line 795: But do we want to stop the entire execution and reject it because one payment failed? I don't think we should as the other payments are processed by the NWC server anyways and that information would be lost.

src/webln/NostrWeblnProvider.ts
```diff
@@ -644,6 +689,139 @@ export class NostrWebLNProvider implements WebLNProvider, Nip07Provider {
...
+            response = JSON.parse(decryptedContent);
+          } catch (e) {
+            console.error(e);
+            clearTimeout(replyTimeoutCheck);
+            sub.unsub();
+            reject({ error: "invalid response", code: "INTERNAL" });
+            return;
+          }
+          if (event.kind == 23195 && response.result) {
+            // console.info("NIP-47 result", response.result);
```
im-adithya, line 764: Should remove this

rolznz:
Update: we need to work on some other things like splitting the NWC client from the WebLN client. I will mark this method as unconfirmed until it is properly added to WebLN and the Alby extension

src/webln/NostrWeblnProvider.ts
rolznz: Update: we should do this, but we have other stuff to work on. There is already a TODO for now.

Issue #182, opened 2024-01-08
Split NWC client and webln provider


bumi:
NWC and webln are not fully compatible. 
currently our NWC client is webln compatible which limits us using some NWC details

We should split this up
and make a NWC client independent of WebLN. 

Maybe we make a NWC class and the NWC webln provider inherits from that one?

rolznz:
Or @bumi 's other suggestion:

> maybe have a NWC client and a NWC WebLN Provider uses that and returns the webln required data.

PR #197, opened 2024-02-04
feat: Split NWC client from NostrWeblnProvider

rolznz:
Fixes https://github.com/getAlby/js-sdk/issues/182

WebLN and NWC are not 100% compatible. This PR splits the two so that the WebLN provider uses a NWC Client (which can also be used on its own as it has more detailed responses and is less restricted than WebLN).

The WebLN provider should be unchanged.

This PR also adds multi_pay_keysend.

TODOs:
- [ ] add examples for all client methods (currently only get-info)
- [ ] re-test multi_methods (currently WIP)
- [ ] review naming (NWCClient class and filename)
- [ ] review exports (nwc)
- [ ] address TODO on Transaction type: `// TODO: review fields (replace with camelCase)`

### DIFF
--- a/src/webln/NostrWeblnProvider.ts
+++ b/src/webln/NostrWeblnProvider.ts
@@ -75,24 +75,42 @@ const nip47ToWeblnRequestMap: Record<Nip47Method, WebLNMethod> = {
 };
 
 export class NostrWebLNProvider implements WebLNProvider, Nip07Provider {
-  private readonly _client: NWCClient;
   private _enabled = false;
+  readonly client: NWCClient;
   readonly subscribers: Record<string, (payload: unknown) => void>;
 
+  /**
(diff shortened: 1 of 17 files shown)


src/webln/NostrWeblnProvider.ts
```diff
@@ -44,106 +34,65 @@ export type ListTransactionsResponse = {
...
 };
 
 export class NostrWebLNProvider implements WebLNProvider, Nip07Provider {
-  relay: Relay;
-  relayUrl: string;
-  secret: string | undefined;
-  walletPubkey: string;
-  options: NostrWebLNOptions;
-  subscribers: Record<string, (payload: unknown) => void>;
+  private readonly _client: NWCClient;
```
rolznz, line 78: make public

PR #204, opened 2024-02-17
Send empty params for `get_info` and `get_balance`

benthecarman:
This should make it more compatible and aligned with the spec


rolznz (approved):
Thanks Ben!

PR #206, opened 2024-02-22
feat: add NWC client method to get wallet service supported methods

rolznz:
As per https://github.com/nostr-protocol/nips/blob/master/47.md#theory-of-operation to get "info" (actually just supported methods) from the relay

### DIFF
--- a/src/NWCClient.ts
+++ b/src/NWCClient.ts
@@ -181,6 +181,35 @@ export class NWCClient {
     return new NWCClient(options);
   }
 
+  static async getWalletServiceSupportedMethods(
+    options?: ConstructorParameters<typeof NWCClient>[0],
+  ): Promise<Nip47Method[]> {
+    const client = NWCClient.withNewSecret(options);
+    await client.relay.connect();
+
+    const events = await client.relay.list(
+      [
+        {
+          kinds: [13194],
+          limit: 1,
+          authors: [client.walletPubkey],
+        },
+      ],
+      {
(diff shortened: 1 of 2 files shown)


src/NWCClient.ts
```diff
@@ -181,6 +181,35 @@ export class NWCClient {
     return new NWCClient(options);
   }
 
+  static async getWalletServiceSupportedMethods(
+    options?: ConstructorParameters<typeof NWCClient>[0],
```
rolznz, line 185: Actually we only need the relay URL here, maybe I should change this to just be a string parameter

rolznz: EDIT: we need to confirm it was the wallet service who posted the event. So I think it's good to keep it like this

src/NWCClient.ts
```diff
@@ -181,6 +181,35 @@ export class NWCClient {
     return new NWCClient(options);
   }
 
+  static async getWalletServiceSupportedMethods(
```
bumi, line 184: how about making on the instance? why do you think it should be on the class level?

```
client = NWCClient.withNewSecret(options);
await client.supportedMethods();
```

bumi (approved):

src/NWCClient.ts
rolznz: That makes sense. Sorry, I originally thought we could request the methods just using the relay URL, but then they can't be filtered by the wallet service pubkey (just in case).

Thanks!

PR #209, opened 2024-02-29
fix: use actual error objects in NWC Client

rolznz:
- [ ] update extension to use Alby JS SDK 3.3.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@getalby/sdk",
-  "version": "3.3.1",
+  "version": "3.3.2",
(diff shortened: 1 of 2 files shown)


bumi:
How about we introduce some more error classes depending on when the error happens.?

e.g. 
Timeout error, 
wallet error,

those inherit from the nip47 error but are a bit more specific. then as a developer I can easily catch the specific type or error

rolznz:
@bumi I've added some more error classes, could you have another look?

bumi (approved):
looks good to me!

PR #210, opened 2024-03-06
chore: try eventemitter3

bumi:
as the events package does not play nicely in a module import setup this is an alternative. 

https://www.npmjs.com/package/eventemitter3

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "prepare": "husky install"
   },
   "dependencies": {
-    "events": "^3.3.0",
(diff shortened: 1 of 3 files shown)


rolznz (approved):
utACK

rolznz:
@pavanjoshi914 any chance you can check this works with the extension, and if so make a patch release?

pavanjoshi914:
looks good! able to receive tokens via event

image

pavanjoshi914 (approved):

PR #222, opened 2024-06-03
Fix: examples for nodeJS 20+

rolznz:
In Node.JS 20+ crypto already exists, and it also does not allow to assign directly to the global object.


bumi (approved):

im-adithya:
tACK

Screenshot 2024-06-03 at 1 54 53 PM

PR #223, opened 2024-06-12
feat: add payment_sent notification

im-adithya:
Adds `payment_sent` support in NIP47 Notifications

### DIFF
--- a/src/NWCClient.ts
+++ b/src/NWCClient.ts
@@ -43,7 +43,8 @@ export type Nip47GetInfoResponse = {
   network: string;
   block_height: number;
   block_hash: string;
-  methods: string[];
+  methods: Nip47Method[];
+  notifications: Nip47NotificationType[];
 };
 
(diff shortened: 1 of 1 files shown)


bumi:
@rolznz what do you think? should we provide the option here on a client level to either subscribe to incoming or outgoing ones? (and just drop the other ones)

rolznz:
@bumi hmm, this is good we decided to implement this now. I guess returning extra notification types could possibly break some clients if they do not specifically check the notification type in their code, and it's also wasteful if they get notifications they don't need.

I wonder if we should add a tag of the notification type. Then clients can use the filter they need in the standard nostr way.

rolznz:
> I wonder if we should add a tag of the notification type. Then clients can use the filter they need in the standard nostr way.

We probably shouldn't do this - the tag will expose extra metadata, so maybe just enable on the client side instead.

bumi (commented):
utACK

PR #224, opened 2024-06-18
chore: add notifications to request map

im-adithya:
get-info shows undefined in between because of this

### DIFF
--- a/src/NWCClient.ts
+++ b/src/NWCClient.ts
@@ -29,7 +29,8 @@ type Nip47SingleMethod =
   | "pay_keysend"
   | "lookup_invoice"
   | "list_transactions"
-  | "sign_message";
+  | "sign_message"
+  | "notifications";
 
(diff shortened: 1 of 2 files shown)


rolznz:
@im-adithya this is not correct, it's not a request.

im-adithya:
Then should I filter out while returning request_methods in get info call? How will the dev know that it supports notifications?

rolznz:
> get-info shows undefined in between because of this

@im-adithya can you show an example?

> Then should I filter out while returning request_methods in get info call?

Why do you need to filter something out?

> How will the dev know that it supports notifications?

can we add the supported notification types to `Nip47GetInfoResponse`?

rolznz:
@im-adithya I think I missed this in the spec PR. It's published as a tag in the replaceable info event but not in the response of the `get_info` NIP-47 request method. Thanks for raising this - I think we need to update that as well.

PR #227, opened 2024-07-01
fix: make wallet service info method return nip-47 methods, not capabilities

rolznz:
Breaking change:

wallet service info method now returns `methods`, not `capabilities`


rolznz:
decided to keep it as capabilities

PR #229, opened 2024-07-08
fix: pass TLV value as hex from nwc webln provider to nwc client keysend

rolznz:
With WebLN keysend TLVs, the value is passed as plaintext:
https://www.webln.guide/building-lightning-apps/webln-reference/webln.keysend

In NIP-47 spec the value should be hex encoded:
https://github.com/nostr-protocol/nips/blob/master/47.md#pay_keysend

This PR fixes the NostrWeblnProvider to pass the TLV values hex-encoded instead of plaintext.

### DIFF
--- a/examples/nwc/client/pay-keysend.js
+++ b/examples/nwc/client/pay-keysend.js
@@ -17,17 +17,38 @@ rl.close();
 const client = new nwc.NWCClient({
   nostrWalletConnectUrl: nwcUrl,
 });
+
+// from https://stackoverflow.com/a/50868276
+const toHexString = (bytes) =>
+  bytes.reduce((str, byte) => str + byte.toString(16).padStart(2, "0"), "");
+
+// Data from https://podcastindex.org/podcast/920666
+const boost = {
+  action: "boost",
+  value_msat: 1000,
+  value_msat_total: 1000,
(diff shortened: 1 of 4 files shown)


im-adithya:
We should also change the new NWC client example to pass hex encoded value to `tlv_records` [here](https://github.com/getAlby/js-sdk/blob/master/examples/nwc/keysend.js) and [here](https://github.com/getAlby/js-sdk/blob/master/examples/nwc/multi-keysend.js)

examples/nwc/client/pay-keysend.js
```diff
@@ -17,17 +17,38 @@ rl.close();
...
+  value_msat: 1000,
+  value_msat_total: 1000,
+  app_name: "⚡ WebLN Demo",
+  app_version: "1.0",
+  feedID: "https://feeds.podcastindex.org/pc20.xml",
+  podcast: "Podcasting 2.0",
+  episode: "Episode 104: A New Dump",
+  ts: 21,
+  name: "⚡ WebLN Demo",
+  sender_name: "Sathoshi Nakamoto",
```
im-adithya, line 37: Sorry but can never get his name wrong ⚡😄

Issue #228, opened 2024-07-02
Update README with NWCClient examples


rolznz:
You can import `{nwc}` from the package and then use `nwc.NWCClient` in the same way as the examples.

PR #230, opened 2024-07-10
chore: add README for nwc client

im-adithya:
Closes https://github.com/getAlby/js-sdk/issues/228

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ or for use without any build tools:
 
 ```html
 <script type="module">
-  import { webln } from "https://esm.sh/@getalby/sdk@3.6.1"; // jsdelivr.net, skypack.dev also work
+  import { nwc } from "https://esm.sh/@getalby/sdk@3.6.1"; // jsdelivr.net, skypack.dev also work
 
   (async () => {
     const client = new nwc.NWCClient({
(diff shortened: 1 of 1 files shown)


rolznz:
@im-adithya this change removes the WebLN provider stuff, which I am not sure is what we want. What I meant is to add a bit of extra documentation on how to use the NWCClient directly

CC @bumi

README.md
```diff
@@ -20,17 +20,25 @@ or for use without any build tools:
 
 ```html
 <script type="module">
-  import { webln } from "https://esm.sh/@getalby/sdk@3.5.1"; // jsdelivr.net, skypack.dev also work
+  import { webln } from "https://esm.sh/@getalby/sdk@3.6.0"; // jsdelivr.net, skypack.dev also work
```
rolznz, line 23: nwc needs to be imported, only webln is imported. And I think we should split this into separate examples

im-adithya:
> this change removes the WebLN provider stuff

This PR only adds NWCClient documentation (and pushes WebLN provider to the bottom) can you check the diffs again?

README.md
```diff
@@ -45,15 +61,175 @@ or for use without any build tools:
...
 ## Nostr Wallet Connect Documentation
 
-Nostr Wallet Connect is an open protocol enabling applications to interact with bitcoin lightning wallets. It allows users to connect their existing wallets to your application allowing developers to easily integrate bitcoin lightning functionality.
+[Nostr Wallet Connect](https://nwc.dev) is an open protocol enabling applications to interact with bitcoin lightning wallets. It allows users to connect their existing wallets to your application allowing developers to easily integrate bitcoin lightning functionality.
 
 The Alby JS SDK allows you to easily integrate Nostr Wallet Connect into any JavaScript based application.
 
+### NWCClient
+
+### Options
```
rolznz, line 70: Can you fix the headers, they should not all be `###`

README.md
```diff
@@ -45,15 +61,175 @@ or for use without any build tools:
...
+import { nwc } from "@getalby/sdk";
+const nwc = new nwc.NWCClient({
+  nostrWalletConnectUrl: loadNWCUrl(),
+}); // loadNWCUrl is some function to get the NWC URL from some (encrypted) storage
+
+// now you can send payments by passing in the invoice in an object
+const response = await nwc.payInvoice({ invoice });
+```
+
+That's it! Unlike WebLN, you don't even have to enable to connect to the relay.
```
rolznz, line 91: Do you think people reading this understand this? and this is not the reason to use or not use WebLN.

rolznz:
When does the user want to use WebLN vs the NWCClient directly, and do they know why? and then we should plan how we can easily communicate that to the user and structure the README

For example, if you're building an app that already uses webln integrated, it makes more sense to use the NostrWebLNClient (Or Bitcoin Connect?)

Alternatively, if you only support NWC (e.g. something like ZapPlanner where you have async payments on your backend) maybe it makes sense to use the NWCClient directly.

rolznz:
Closing in favor of https://github.com/getAlby/js-sdk/pull/344 which splits up the documentation

PR #232, opened 2024-07-27
feat: add metadata field to make_invoice

rolznz:
See https://github.com/getAlby/hub/pull/357


src/NWCClient.ts
```diff
@@ -131,6 +131,7 @@ export type Nip47MakeInvoiceRequest = {
   description?: string;
   description_hash?: string;
   expiry?: number; // in seconds
+  metadata?: unknown;
```
rolznz, line 134:
```suggestion
  metadata?: unknown; // TODO: update to also include known keys (payerData, nostr, comment)
```

PR #253, opened 2024-09-10
fix: add custom error for nwc relay connection

rolznz:


rolznz:
@bumi it's done automatically by the JS SDK. Consumers do not need to do anything. But more is done in nostr-tools in version 2.X, which we still need to update to. Then we can remove some of the code from this codebase.

bumi (approved):

Issue #261, opened 2024-10-08
Implement nwc get_budget method


rolznz:
See https://github.com/getAlby/hub/pull/726

PR #264, opened 2024-10-19
feat: add get_budget to NWCClient

rolznz:
Fixes https://github.com/getAlby/js-sdk/issues/261

### DIFF
--- a/src/NWCClient.ts
+++ b/src/NWCClient.ts
@@ -24,6 +24,7 @@ type WithOptionalId = {
 type Nip47SingleMethod =
   | "get_info"
   | "get_balance"
+  | "get_budget"
   | "make_invoice"
   | "pay_invoice"
   | "pay_keysend"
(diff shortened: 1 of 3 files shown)


src/NWCClient.ts
```diff
@@ -47,6 +48,16 @@ export type Nip47GetInfoResponse = {
   notifications?: Nip47NotificationType[];
 };
 
+export type Nip47GetBudgetResponse =
+  | {
+      used_budget: number; // msats
+      total_budget: number; // msats
+      renews_at?: number; // timestamp
```
im-adithya, line 55: This is defined as `number | undefined` but if `renewal_period` is never we get `renews_at: null` 

Screenshot 2024-10-22 at 1 20 02 PM

im-adithya (approved):
Just a small nit, otherwise tACK

src/NWCClient.ts
rolznz: Good catch, I think we should fix this in Alby Hub to omit `renews_at` in the json.

rolznz: Fixed here: https://github.com/getAlby/hub/pull/726/commits/634e54d778bee73c0c4f73dceffa7d05b7a1d21c

im-adithya: Resolved by https://github.com/getAlby/hub/pull/726/commits/634e54d778bee73c0c4f73dceffa7d05b7a1d21c

PR #270, opened 2024-11-27
feat: allow passing metadata in nwc pay_invoice and multi_pay_invoice

rolznz:
Alby Hub can store this data, enabling interesting applications to be developed, and finally enable a nice way to store payment metadata.

### DIFF
--- a/src/NWCClient.ts
+++ b/src/NWCClient.ts
@@ -127,6 +127,7 @@ export type Nip47Notification =
 
 export type Nip47PayInvoiceRequest = {
   invoice: string;
+  metadata?: unknown;
   amount?: number; // msats
 };
 


src/NWCClient.ts
```diff
@@ -127,6 +127,7 @@ export type Nip47Notification =
 
 export type Nip47PayInvoiceRequest = {
   invoice: string;
+  metadata?: unknown;
```
pavanjoshi914, line 130: instead of unknown couldn't be an empty object or something?

rolznz: @pavanjoshi914 an empty object won't work, it will create issues when you actually try to pass an object with properties set.

PR #273, opened 2024-11-29
feat: add nip 44 and versioning support

im-adithya:
Ready for review

- This needs the long pending update to `nostr-tools` v2 as nip44 library doesn't match the latest spec (and hence gives errors as it doesn't match with the Hub's nip-44)
- For now I added the `nip-44` folder so we can test it, in the meantime I'll work on a PR to shift to `nostr-tools` v2 so we can remove it.


src/NWCClient.ts
```diff
@@ -895,6 +927,7 @@ export class NWCClient {
     resultValidator: (result: T) => boolean,
   ): Promise<(T & { dTag: string })[]> {
     await this._checkConnected();
+    await this._checkCompatibility();
```
rolznz, line 930: why did you choose to check compatibility here?

what about once in the constructor? then we also know what notification kind to subscribe to?

src/NWCClient.ts
```diff
@@ -1048,4 +1085,59 @@ export class NWCClient {
...
+        throw new Nip47UnsupportedVersionError(
+          `no compatible version found between wallet and client`,
+          "UNSUPPORTED_VERSION",
+        );
+      }
+      this.version = compatibleVersion;
+    }
+  }
+
+  private selectHighestCompatibleVersion(
```
rolznz, line 1105: It is nice, but I am wondering if such a complex algorithm is needed?

What about just:
if the wallet supports version 1.0, use it. Otherwise use 0.0.

As Bumi mentioned:
> I do not expect many versions (and rather bigger changes like the switch to nip-44) because too many versions (of a spec) lead to incompatibilities which must be avoided.

rolznz:
> This needs the long pending update to nostr-tools v2

@im-adithya this is great! :rocket: 

please let me know when it's ready for testing.

src/NWCClient.ts
```diff
@@ -687,14 +715,14 @@ export class NWCClient {
...
-    let sub: Sub<23196> | undefined;
+    let sub: Sub<number> | undefined;
     (async () => {
       while (subscribed) {
         try {
           await this._checkConnected();
           sub = this.relay.sub([
             {
-              kinds: [23196],
+              kinds: [...(this.supportedVersion ? [23196] : [23197])],
```
im-adithya, line 725: This is wrong, if supported version exists and is "0.0", it should be 23196, not 23197

src/NWCClient.ts
im-adithya: It's an async function (fetching wallet service info), so I thought of doing the call before making a request

> then we also know what notification kind to subscribe to?

We are doing that right? We store the version in `this.version`/`this.supportedVersion` and use that before passing the kind in `subscribeNotifications`

src/NWCClient.ts
im-adithya: Hmmm, I was also thinking about this, but then thought to add this as this is an sdk and can face any possibility... (a weird case where the wallet service info gives "0.5" or "1.1" for example)

im-adithya:
Done, ready for testing!

rolznz:
I wonder if it's good to add the NIP-47 version to the app table, and default to 0.0 for existing entries. We do not republish info events for apps on startup. Later we might want to show which connections are using NIP-04 so that the user can remove them and re-add them? CC @reneaaron @frnandu @bumi

src/NWCClient.ts
```diff
@@ -1084,4 +1122,59 @@ export class NWCClient {
...
+      const compatibleVersion = this.selectHighestCompatibleVersion(
+        walletServiceInfo.versions,
+      );
+      if (!compatibleVersion) {
+        throw new Nip47UnsupportedVersionError(
+          `no compatible version found between wallet and client`,
+          "UNSUPPORTED_VERSION",
+        );
+      }
+      this.version = compatibleVersion;
```
rolznz, line 1138: if the version is 0.0 should we log a warning?

src/NWCClient.ts
rolznz: if we leave it as-is, please add a few tests to cover it

im-adithya: Added!

src/NWCClient.ts
im-adithya: Yeah I think it's good to add that, done!

src/NWCClient.ts
rolznz: @im-adithya I think the same comment applies here. We should be able to check if the wallet service supports 1.0, then use it, otherwise use 0.0

src/NWCClient.test.ts
```diff
@@ -56,3 +56,85 @@ describe("NWCClient", () => {
...
+describe("selectHighestCompatibleVersion", () => {
+  let nwcClient: NWCClient;
+  let selectVersion: (walletVersions: string[]) => string | null;
+  const ORIGINAL_SUPPORTED_VERSIONS = NWCClient.SUPPORTED_VERSIONS;
+
+  beforeEach(() => {
+    nwcClient = new NWCClient();
+    // Access the private method using type assertion
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    selectVersion = (nwcClient as any).selectHighestCompatibleVersion;
```
rolznz, line 69: Does `selectHighestCompatibleVersion` have to be an instance method? if we moved it out then it can be more easily tested. As far as I can see it can be easily moved out

src/NWCClient.test.ts
```diff
@@ -56,3 +56,85 @@ describe("NWCClient", () => {
...
+  });
+
+  afterEach(() => {
+    // Restore the original SUPPORTED_VERSIONS
```
rolznz, line 73: This feels a bit hacky to me. Maybe can be solved by the above comment. 

Also, the JS SDK knows what versions it supports, I think editing this value and testing it does not really make sense

src/NWCClient.ts
rolznz: Sorry, I was wrong about just adding tests. I talked with Bumi and we can see here how much extra complexity and testing we have to do, which is not really necessary right now since we know there will probably only be two versions until something major again comes along, which we have no idea when that will be

src/NWCClient.test.ts
im-adithya: We agreed to not have tests given `selectHighestCompatibleVersion` is now hard-coded/simplified

src/NWCClient.ts
im-adithya: Done!

rolznz (approved):
tACK

PR #275, opened 2024-12-03
chore(deps): update nostr-tools package

im-adithya:
## Description

This updates the nostr-tools package to 2.9.4 (the latest is 2.10.4 but it has a timeout leak, opened a [PR](https://github.com/nbd-wtf/nostr-tools/pull/461) to fix it there)

### DIFF
--- a/src/NWCClient.ts
+++ b/src/NWCClient.ts
@@ -1,17 +1,17 @@
 import {
   nip04,
-  relayInit,
-  getEventHash,
   nip19,
-  generatePrivateKey,
+  finalizeEvent,
+  generateSecretKey,
+  getEventHash,
   getPublicKey,
-  Relay,
   Event,
-  UnsignedEvent,
-  finishEvent,
-  Sub,
+  EventTemplate,
+  Relay,
 } from "nostr-tools";
 import { NWCAuthorizationUrlOptions } from "./types";
+import { hexToBytes, bytesToHex } from "@noble/hashes/utils";
+import { Subscription } from "nostr-tools/lib/types/abstract-relay";
 
 type WithDTag = {
   dTag: string;
@@ -252,7 +252,7 @@ export class NWCClient {
 
   static withNewSecret(options?: ConstructorParameters<typeof NWCClient>[0]) {
     options = options || {};
-    options.secret = generatePrivateKey();
+    options.secret = bytesToHex(generateSecretKey());
     return new NWCClient(options);
   }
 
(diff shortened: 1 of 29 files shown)


bumi:
yay

src/NWCClient.ts
```diff
@@ -459,19 +459,31 @@ export class NWCClient {
...
+
+      sub.onevent = (event) => {
+        events.push(event);
+      };
+
+      sub.oneose = () => {
+        sub.close();
+        resolve(events);
+      };
+    });
```
bumi, line 486: the behavior here is the same? we subscribe and on oneose the promise will be resolved? 
what's after the timeout?

bumi (commented):
when do we actually need those `eoseTimeout` on the subscriptions?

src/NWCClient.ts
im-adithya: If the timeout is exceeded, the oneose function is called, which closes the subscription and resolves with events (`[]`)

Timeout function is added on firing in new relay: https://github.com/nbd-wtf/nostr-tools/blob/9c50b2c65515c6282872cb2cd8b4a58f7981c269/abstract-relay.ts#L349-L361

Old list method: https://github.com/nbd-wtf/nostr-tools/blob/318e3f8c8867cb183684622cdedf6a9ee49cee78/relay.ts#L305-L321

im-adithya: > when do we actually need those eoseTimeout on the subscriptions?

I'm assuming when there are a lot of stored events that we keep fetching till the last event is reached, but in our case it's useless as we only have one persistent 13194 info event

rolznz: Right now it takes 10 seconds for the method to return, so I think if we get an event we should unsubscribe

bumi: I did not run the code currently, but does it take 10 seconds? 
Shouldn't the relay send a eose and then the that function should fire?

im-adithya: I checked the previous code and surprisingly the `eoseSubTimeout` wasn't even being used

> Right now it takes 10 seconds for the method to return

It returned almost instantaneously for me, it returns as soon as eose is reached no?

rolznz: True, that's right.

It's only 10 seconds in React Native in the latest version of Expo, (because of another issue which is also causing publishing to never return). I found a way to fix it in `nostr-tools` and will add a PR there.

rolznz (approved):
tACK

PR #297, opened 2025-01-18
feat: new NWC deeplink flow to support other relays and wallet pubkeys

rolznz:
This updated the new client-initiated auth flow to be independent of relays and wallet service pubkeys, by instead receiving the nostr wallet connect string (everything apart from the secret, which is generated locally - and only the pubkey is passed to the auth URL)

**Breaking changes**

Old `initNWC()` and `withNewSecret` are removed.

See changes required in Alby Hub: https://github.com/getAlby/hub/pull/1009


src/NWCClient.ts
```diff
@@ -456,7 +457,10 @@ export class NWCClient {
       };
 
       const onMessage = (message: {
-        data?: { type: "nwc:success" | unknown };
+        data?: {
+          type: "nwc:success" | unknown;
+          nostrWalletConnectUrl?: string;
```
rolznz, line 462: this needs to be changed to `relayUrl` and `walletPubkey` to be clear that the app secret is client-generated.

Dayvvo (commented):
Hey @rolznz !! . gone through the code and tested this PR with bitcoin-connect locally. Looks good!

However I have one question. It seems there is no way to call the fromAuthorizationUrl from the `  NostrWebLNProvider  ` class. Only when using the `  NwcClient  ` class directly.  currently some flows in bitcoin-connect use the ` 
 NostrWebLNProviderClass  `, as is suggested for a web app flow. The only way I could test this locally was to use the NwcClient class directly. Is this intentional?

rolznz:
@Dayvvo awesome! You're right, I see this is an issue since the NostrWeblnProvider then would have to initialize a new NWCClient. I think we need to fix this so we can optionally pass a NWCClient into the NostrWeblnProvider constructor. We could also have a `fromAuthorizationUrl` method on the NostrWeblnProvider just for convenience that creates a NWCClient and then returns a new NostrWeblnProvider from that NWCClient.

I will try fix this before this PR is merged, thanks!

PR #299, opened 2025-01-20
feat: add custom-timeout-values-for-requester-method

### LINKED ISSUE #295: Reduce 1 minute replyTimeout when the "get_info" event is called from the executeMultiNip47Request method in NWCClient.ts
### Problem definition:

I noticed this issue when trying to connect wallets in my web app using [bitcoin-connect](https://github.com/getAlby/bitcoin-connect). There were inconsistencies in the wait time when connecting a wallet as sometimes the wallet would load for a full minute which just seemed like bad UX for an end user.  I traced this long wait time to the 1 minute replyTimeout found in the `executeMultiNip47Request` method when the `getInfo` method found in [NWCClient.ts](https://github.com/getAlby/js-sdk/blob/943acf3327316173dfae0f2e248dbd2013ff3274/src/NWCClient.ts) was called.

### Fix
The 60 seconds timeout add an optional argument for timeout to the `executeMultiNip47Request` method. This allows `getInfo` to pass a shorter timeout duration when it is called. This would allow for a longer timeout when needed for other events like `pay_invoice` and a shorter timeout when it is needed

Dayvvo:
fixes #295

### DIFF
--- a/src/NWCClient.ts
+++ b/src/NWCClient.ts
@@ -67,6 +67,11 @@ export type Nip47PayResponse = {
   preimage: string;
 };
 
+type Nip47TimeoutValues = {
+  replyTimeout?: number;
+  publishTimeout?: number;
+};
+
 export type Nip47MultiPayInvoiceRequest = {
   invoices: (Nip47PayInvoiceRequest & WithOptionalId)[];
 };
@@ -544,6 +549,7 @@ export class NWCClient {
         "get_info",
         {},
         (result) => !!result.methods,
+        { replyTimeout: 10000 },
       );
       return result;
     } catch (error) {
@@ -819,6 +825,7 @@ export class NWCClient {
     nip47Method: Nip47SingleMethod,
     params: unknown,
     resultValidator: (result: T) => boolean,
+    timeoutValues?: Nip47TimeoutValues,
   ): Promise<T> {
     await this._checkConnected();
     await this._checkCompatibility();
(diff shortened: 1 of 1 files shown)


src/NWCClient.ts
```diff
@@ -819,6 +825,7 @@ export class NWCClient {
     nip47Method: Nip47SingleMethod,
     params: unknown,
     resultValidator: (result: T) => boolean,
+    timeoutValues?: Nip47TimeoutValues,
```
rolznz, line 828: Less important, but can this also be added to the `executeMultiNip47Request` method?

src/NWCClient.ts
```diff
@@ -544,6 +549,7 @@ export class NWCClient {
         "get_info",
         {},
         (result) => !!result.methods,
+        { replyTimeout: 10000 },
```
rolznz, line 552: Originally you also wanted to be able to customize this on a per-request basis right? but I guess this is ok.

I would probably also apply this smaller timeout to other non-payment requests (get balance, get budget, list transactions, etc.)

Dayvvo: @rolznz done. I thought of making a shorter reply timeout the default in the Nip47Request method but it seems there are as much payment related requests as non-payment. So this is good

src/NWCClient.ts
rolznz: @Dayvvo what do you think about this one?

rolznz (approved):
utACK

src/NWCClient.ts
rolznz: Thanks!

src/NWCClient.ts
Dayvvo: @rolznz  sounds good. Made a commit for that !!

PR #308, opened 2025-02-13
feat: custom EventEmitter classes

pavanjoshi914:
remove emittery package causing build issues
image

### DIFF
--- a/src/eventEmitter/EventEmitter.ts
+++ b/src/eventEmitter/EventEmitter.ts
@@ -1,8 +1,7 @@
-import { EventName, EventListener, Token } from "../types";
+import { EventListener, EventName } from "../types";
 
 export class EventEmitter {
-  // eslint-disable-next-line @typescript-eslint/ban-types
-  private events: { [key: string]: Function[] } = {};
+  private events: { [key: string]: EventListener[] } = {};
 
   on(event: EventName, listener: EventListener) {
     if (!this.events[event]) {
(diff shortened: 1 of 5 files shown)


src/eventEmitter/EventEmitter.ts
```diff
@@ -0,0 +1,23 @@
...
+    }
+    this.events[event].push(listener);
+  }
+
+  off(event: EventName, listener: EventListener) {
+    if (!this.events[event]) return;
+    this.events[event] = this.events[event].filter((l) => l !== listener);
+  }
+
+  emit(event: EventName, payload: Token | Error) {
```
rolznz, line 19: is the payload type right here? or can it just be `unknown`?

src/eventEmitter/EventEmitter.ts
```diff
@@ -0,0 +1,23 @@
+import { EventName, EventListener, Token } from "../types";
+
+export class EventEmitter {
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
rolznz, line 4: why is this eslint comment needed? Is `Function` correct for the type?

rolznz: should it be `(payload: unknown) => void`?

rolznz (approved):
utACK

src/eventEmitter/EventEmitter.ts
pavanjoshi914: ``` this._tokenEvents.emit("tokenRefreshFailed", error as Error);```
```this._tokenEvents.emit("tokenRefreshed", this.token);```

there are only two instance of emit. where payload is either Token or Error

src/eventEmitter/EventEmitter.ts
pavanjoshi914: i changed it to EventListener[] and payload to any

bumi (approved):
utACK

bumi:
great to get rid of that dependency!

PR #309, opened 2025-02-13
chore: switch to encryption tag from versioning

im-adithya:
Implements https://github.com/nostr-protocol/nips/pull/1780 instead of versioning


src/NWCClient.ts
```diff
@@ -1125,33 +1164,46 @@ export class NWCClient {
   }
 
   private async _checkCompatibility() {
```
im-adithya, line 1166: Minor: I realised that we only check the encryption/version tag from the info event and use the same while doing the request but we don't really check it in the response, but I think it should be okay

src/NWCClient.ts
```diff
@@ -235,9 +235,10 @@ export class NWCClient {
   lud16: string | undefined;
   walletPubkey: string;
   options: NWCOptions;
-  version: string | undefined;
+  encryptionOrVersion: string | undefined;
```
rolznz, line 238: can we just use `encryption` - this will otherwise be hard to remove later. We do not need to know the version actually.

src/NWCClient.ts
```diff
@@ -235,9 +235,10 @@ export class NWCClient {
...
 
   static SUPPORTED_VERSIONS = ["0.0", "1.0"];
+  static SUPPORTED_ENCRYPTIONS = ["nip04", "nip44_v2"];
```
rolznz, line 241: can you add a TODO to the version code `// TODO: remove after 2026-01-01` or something

src/NWCClient.ts
```diff
@@ -334,11 +335,21 @@ export class NWCClient {
...
-      throw new Error("Missing version");
+  get supportedEncryptionOrVersion(): string {
+    if (!this.encryptionOrVersion) {
+      throw new Error("Missing encryption or version");
     }
-    return this.version;
+    return this.encryptionOrVersion;
+  }
+
+  isUsingVersioning(encryptionOrVersion = this.encryptionOrVersion): boolean {
```
rolznz, line 345: code like this will be unfortunately hard to remove, let's huddle together about it next week

im-adithya:
Ready to merge cc @rolznz

rolznz (approved):
tACK

PR #329, opened 2025-03-01
Chore: nwc encryption feedback

rolznz:
Minor code cleanup, renaming methods, + one fix for the legacy version code ("nip44" should be "nip44_v2")


im-adithya (approved):
tACK

PR #331, opened 2025-03-03
fix: re-subscribe after relay disconnections in NWAClient

rolznz:
Code is copied from NWCClient `subscribeNotifications` (which seems to be working well there). In this case I think it's more complicated to try extracting the duplicated code, maybe we can revisit this if we have a third case where we subscribe like this.

Tested with Zappy Bird on iOS, plus locally with shutting down and restarting local relay

### DIFF
--- a/src/NWAClient.ts
+++ b/src/NWAClient.ts
@@ -7,6 +7,7 @@ import {
   Nip47NotificationType,
   NWCClient,
 } from "./NWCClient";
+import { Subscription } from "nostr-tools/lib/types/abstract-relay";
 
 export type NWAOptions = {
   relayUrl: string;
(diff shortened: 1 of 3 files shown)


src/NWAClient.ts
```diff
@@ -170,47 +171,87 @@ export class NWAClient {
...
+              endPromise?.();
+            };
+            this.relay.onclose = onRelayDisconnect;
+          });
+          if (onRelayDisconnect !== undefined) {
+            this.relay.onclose = null;
+          }
+        } catch (error) {
+          console.error(
+            "error subscribing to notifications",
```
im-adithya, line 236: c&p error, also in the comment below

rolznz: oops, thanks!

Issue #343, opened 2025-03-18
Simplify fiat amount usage


bumi:
Using fiat amounts is still a bit complicated. 

maybe we can simpify this:

```js
const amount = await fiat.getSatoshiValue({
  amount: 0.1,
  currency: "usd",
});

const invoice = await nwcClient.makeInvoice({
  amount: amount * 1000,
  description: "Video access",
});

```

```js
const invoice = await nwcClient.makeInvoice({
  amount: USD(0.1),
  description: "Video access",
});

`USD` is an object that can return sats and millisats values and the SDK checks if it gets a number or an object that has a `toMilliSats()` function or something.

Issue #342, opened 2025-03-17
Implement `verifyPayment` on the makeInvoice return value


bumi:
this makes it easy to check if the invoice got paid. 

How can we make this easier? 

```js
const nwcClient = new nwc.NWCClient({
  nostrWalletConnectUrl: NOSTR_WALLET_URL,
});

const amount = await fiat.getSatoshiValue({
  amount: 0.1,
  currency: "usd",
});

const invoice = await nwcClient.makeInvoice({
  amount: amount * 1000,
  description: "Video access",
});

const lookup = await nwcClient.lookupInvoice({
  invoice: invoice.invoice,
});
```

instead of another manual nwcClient call to use `lookupInvoice` we cold have a `verifyPayment` or just a `isPaid` function that checks this.

```js
const invoice = await nwcClient.makeInvoice({
  amount: amount * 1000,
  description: "Video access",
});

await invoice.isPaid();

```

maybe we can also add a callback and subscribe to the notifications?

```
invoice.onPaid(() => {
  giveAccess();
})

**Comment by rolznz** [comment]
@bumi could you add a bit more info? is this for the OAuth implementation?

For NWC you can simply use `lookup_invoice` to check the status. But I don't think this is possible with the OAuth client / Oauth webln provider

**Comment by bumi** [comment]
In combination with bitcoin-connect this could then allow code like: 


```js
const nwcClient = new nwc.NWCClient({
  nostrWalletConnectUrl: NOSTR_WALLET_URL,
});

const invoice = await nwcClient.makeInvoice({
  amount: USD(0.1),
  description: "Video access",
});

launchPaymentModal({invoice });

// maybe launchPaymentModal() could use this `onPaid` function on the `invoice` itself.
// otherwise: 

const {setPaid} = launchPaymentModal({invoice: invoice.invoice });

invoice.onPaid(setPaid)

```

rolznz:
Linking also to https://github.com/getAlby/js-sdk/issues/343

We could have a new "Layer" or simple version for starters to try out the JS SDK. For more custom usage, they can use the NWCClient directly. Also see how it combines with bitcoin connect.

bumi:
somehow I like to have it on the `NWCClient` to not introduce another object or way of doing things. But not using `makeInvoice` there is a good idea I think. have the `makeInvoice` a raw call and have a different function a bit more high level 

```js

const nwcClient = new nwc.NWCClient({
  nostrWalletConnectUrl: NOSTR_WALLET_URL,
[...]

PR #344, opened 2025-03-20
feat: add simpler LNClient class

rolznz:
Closes https://github.com/getAlby/js-sdk/issues/343 and https://github.com/getAlby/js-sdk/issues/342

This PR adds a simpler way for new developers to interact with a lightning wallet. You only need a couple lines to do cool stuff. Unlike the `NWCClient`, the `LNClient` uses satoshi values (is this confusing?)

Alby JS SDK is now dependent on the lightning tools package. (Pros: devs don't need to install both and could have some of the functions easily available here)

Most of the README is now moved into separate readme files (see docs)

Separate components (NWC, WebLN, LNClient, OAuth) are now in separate folders. This is an update to 5.0.0 as some of the imports are breaking changes.

OAuth examples are now moved into a subfolder.

I haven't proxied all the methods yet, I wonder if it's better to just make the internal nwcClient publicly accesible (as it is now)

TODOs:
- [x] address feedback
- [ ] move fiat stuff to lightning tools (https://github.com/getAlby/js-lightning-tools/pull/202) - _decided to postpone for now_
- [x] add documentation for NWC Wallet Service


src/LNClient.ts
```diff
@@ -0,0 +1,100 @@
...
+      this.nwcClient = credentials;
+    } else {
+      this.nwcClient = new NWCClient(credentials);
+    }
+  }
+
+  pay(invoice: string, args?: Omit<Nip47PayInvoiceRequest, "invoice">) {
+    return this.nwcClient.payInvoice({
+      invoice,
+      ...(args || {}),
```
rolznz, line 49: amount needs to be fixed here or removed (sat - millisat issue - and only required for 0-amount invoices)

rolznz: EDIT: also for lightning addresses, I need to update this

docs/ln.md
```diff
@@ -0,0 +1,21 @@
+# Lightning Network Client (LNClient) Documentation
+
+The LNClient is a high level wrapper around the [NWCClient](./nwc.md) which helps you easily get started interacting with the lightning network.
+
+For example, to make a payment:
+
+```js
+await new LNClient(credentials).pay(invoice);
```
rolznz, line 8:
```suggestion
await new LNClient(credentials).pay("lnbc..."); // pay a lightning invoice
```

rolznz: Also add:
```
// ... or pay to a lightning address:
await new LNClient(credentials).pay("hello@getalby.com");
```

docs/ln.md
```diff
@@ -0,0 +1,21 @@
+# Lightning Network Client (LNClient) Documentation
+
+The LNClient is a high level wrapper around the [NWCClient](./nwc.md) which helps you easily get started interacting with the lightning network.
```
rolznz, line 3:
```suggestion
The LNClient helps you easily get started interacting with the lightning network. It is a high level wrapper around the [NWCClient](./nwc.md) which is compatible with many different lightning wallets.
```

src/LNClient.ts
```diff
@@ -0,0 +1,125 @@
...
+export const GBP = (amount: number) =>
+  ({ amount, currency: "GBP" }) satisfies FiatAmount;
+export const CHF = (amount: number) =>
+  ({ amount, currency: "CHF" }) satisfies FiatAmount;
+
+export class LNClient {
+  readonly nwcClient: NWCClient;
+
+  constructor(credentials: LNClientCredentials) {
+    if (typeof credentials === "string") {
```
bumi, line 35: I like that it supports both!

src/LNClient.ts
```diff
@@ -0,0 +1,125 @@
...
+        const onNotification = (notification: Nip47Notification) => {
+          if (
+            notification.notification.payment_hash === transaction.payment_hash
+          ) {
+            unsubscribeFunc();
+            callback(notification.notification);
+          }
+        };
+
+        unsubscribeFunc = await this.nwcClient.subscribeNotifications(
```
bumi, line 98: what happens if I call this too late? 

shouldn't we return a class that automatically subscribes to the notifications (and if not supported polls the status) and then also unsubscribes again after the payment is received?

src/LNClient.ts
```diff
@@ -0,0 +1,100 @@
...
+export const EUR = (amount: number) =>
+  ({ amount, currency: "EUR" }) satisfies FiatAmount;
+export const JPY = (amount: number) =>
+  ({ amount, currency: "JPY" }) satisfies FiatAmount;
+export const GBP = (amount: number) =>
+  ({ amount, currency: "GBP" }) satisfies FiatAmount;
+export const CHF = (amount: number) =>
+  ({ amount, currency: "CHF" }) satisfies FiatAmount;
+
+export class LNClient {
```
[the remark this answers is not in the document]
rolznz: What do you mean? that with this naming it doesn't support on-chain? or that the name should be shortened to `LN`?

src/LNClient.ts
```diff
@@ -0,0 +1,100 @@
...
+
+type LNClientCredentials = string | NWCClient | NewNWCClientOptions;
+type FiatAmount = { amount: number; currency: string };
+
+/**
+ * An amount in satoshis, or an amount in a fiat currency
+ */
+type Amount = number | FiatAmount;
+
+// Most popular fiat currencies
```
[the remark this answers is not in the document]
rolznz: Ah cool, then the `LNClient` class doesn't need to do parsing or fetching or depend on `fiat` if it can always get `millisats` from the `Amount` class.

`USD(1)` would need to return a promise to an amount object, I need to see how we handle that inside `LNClient`.

src/LNClient.ts
rolznz: It's no issue if we call it too late. Nice ideas. I will change this and put the class in a separate file

src/LNClient.ts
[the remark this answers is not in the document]
bumi: we could define the interface that we check if there is a `millisats` property available and use this. 

```
const amount = await USD(1.21);
pay("foo@bar.com", amount);
```

and we just check if amount is a number or if it has a millisats property. It's the job of the `USD` to know about the millisats property.

src/LNClient.ts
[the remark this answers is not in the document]
bumi: was just thinking about a name and wondering if `client` is needed. - we have the NWC client how to not confuse it with LN client? (as both do also LN stuff)

so I thought maybe just keep it small and short.... `LN` :)

src/LNClient.ts
rolznz: I've extracted this to a new class called `ReceiveInvoice` (not sure that's the best name but all its logic is specific to receiving)

src/LNClient.ts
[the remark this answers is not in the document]
rolznz: Update: I didn't add support for number directly, I'm worried about using the incorrect unit (millisats or sats)

examples/lnclient/paywall.js
```diff
@@ -0,0 +1,34 @@
...
+
+const nwcUrl =
+  process.env.NWC_URL ||
+  (await rl.question("Nostr Wallet Connect URL (nostr+walletconnect://...): "));
+rl.close();
+
+const client = new LN(nwcUrl);
+const request = await client.receive(USD(1.0));
+
+qrcode.generate(request.invoice, { small: true });
```
bumi, line 19:
```suggestion
qrcode.generate(request.invoice.paymentRequest,, { small: true });
console.info(request.invoice.paymentRequest);
```

examples/lnclient/paywall.js
```diff
@@ -0,0 +1,34 @@
...
+console.info("Waiting for payment...");
+
+const unsub = await request.onPaid(() => {
+  console.info("received payment!");
+  client.close();
```
bumi, line 24: I am wondering if we always should automatically close the client connection, so that the user of the LN client does not need to care about it? - it will be forgotten anyway.

src/lnclient/LNClient.ts
```diff
@@ -0,0 +1,94 @@
...
+   * Make a payment
+   * @param recipient a BOLT-11 invoice or lightning address
+   * @param amount the amount to pay, only required if paying to a lightning address or the amount is not specified in the BOLT 11 invoice.
+   * @param args additional options, e.g. to store metadata on the payment
+   * @returns the receipt of the payment, and details of the paid invoice.
+   */
+  async pay(
+    recipient: string,
+    amount?: Amount,
+    args?: Omit<Nip47PayInvoiceRequest, "invoice" | "amount">,
```
bumi, line 41:
```suggestion
    recipient: string,
    amount?: Amount,
    comment?: string,
    args?: Omit<Nip47PayInvoiceRequest, "invoice" | "amount">,
```

do you think it makes sense to get the comment in there which we pass on when requesting the invoice from the LN address? 
it's often used to add a message.

bumi (approved):
this is great!

examples/lnclient/paywall.js
rolznz: it will unsubscribe automatically if the invoice is paid. This is only if e.g. the user decided they do not want to pay (but I agree it will be forgotten/unused - but still I'm hesitant to remove it entirely because it creates a memory leak)

bumi:
maybe some more Money inspiration: https://www.npmjs.com/package/ts-money (not for here as a dependency though)

examples/lnclient/paywall.js
bumi: but if it unsubscribes automatically then we don't need it here in the onPaid callback?

src/lnclient/FiatAmount.ts
```diff
@@ -0,0 +1,19 @@
...
+  constructor(amount: number, currency: string) {
+    this.satoshi = fiat.getSatoshiValue({
+      amount,
+      currency,
+    });
+  }
+}
+
+// Most popular fiat currencies
+export const USD = (amount: number) => new FiatAmount(amount, "USD");
```
rolznz, line 15: These are currently exported directly which muddies the root level exports. In js-lightning-tools they can be exported under `fiat`

examples/lnclient/paywall.js
rolznz: Only if we have another way of unsubscribing if the invoice is not paid (like a timeout or something?)

bumi: sorry, I don't fully understand.
we have the unsubscribe from the subscription and closing the client connection.

which one is needed?

on this example you call `client.close();` in the `onPaid` callback. Can we get rid of this? and always open/close the connection automatically if needed?

do we also need a `onTimeout(5, () => {})`  helper or something?

rolznz: > which one is needed?
Neither the `close();` or the `unsubscribe()` are really needed in the example since the process exits anyway. I could remove them from the examples.

> on this example you call client.close(); in the onPaid callback. Can we get rid of this? and always open/close the connection automatically if needed?

This would cause slow down (disconnecting and re-connecting the websocket) in the case where a NWCClient is used multiple times, and also if there's a subscription active it becomes complicated whether to know if the connection should be closed or not.

> do we also need a onTimeout(5, () => {}) helper or something?

This would be cool, yeah - although should we also set the invoice expiry to be in sync? that way there is no worry that the payment is later made but we did not see it

src/lnclient/LNClient.ts
rolznz: I'm not sure about putting it as its own argument since it only applies to payments to a LN address.

To be consistent with how we handle metadata for received payments in the hub, I think setting args as `{metadata: {comment, payer_data}}` would do it, I just need to update the typings here so it's more obvious for users.

Related PR: https://github.com/getAlby/go/pull/247

rolznz: https://github.com/getAlby/js-sdk/pull/356

README.md
```diff
@@ -26,525 +28,58 @@ or for use without any build tools:
...
-  nostrWalletConnectUrl:
-    "nostr+walletconnect://69effe7b49a6dd5cf525bd0905917a5005ffe480b58eeb8e861418cf3ae760d9?relay=wss://nostr.bitcoiner.social&secret=c60320b3ecb6c15557510d1518ef41194e9f9337c82621ddef3f979f668bfebd",
-}); // use defaults
-await nwc.enable(); // connect to the relay
-const response = await nwc.sendPayment(invoice);
-console.log(response.preimage);
-
-nwc.close(); // close the websocket connection
+import { LN, USD } from "@getalby/sdk";
+await new LN(credentials).pay("lnbc..."); // pay a lightning invoice
```
rolznz, line 41:
```suggestion
const credentials = "nostr+walletconnect://..."; // the NWC connection credentials
await new LN(credentials).pay("lnbc..."); // pay a lightning invoice
```

examples/lnclient/paywall.js
bumi: > although should we also set the invoice expiry to be in sync? 

hmm, an invoice can be valid for 24h for example, but a typical app would only await a few minutes and then it is pretty sure that the user will no longer pay. 
imo this is some application level logic and I would leave it up to the app. 
a `onTimeout` function imo would also be more of syntax sugar and to answer some default flow maybe? e.g. when I do a paywall I always want to handle when the payment is received or when the user does not pay (e.g. show an error or similar)

PR #346, opened 2025-03-29
feat: add NWC Wallet Service

rolznz:
This makes it much easier for wallets to add support for NWC, by not having to implement the lower-level NWC spec such as encrypting/decrypting and publishing events, subscribing to the relay etc.

TODOs:
- [ ] add documentation
- [x] re-connection seems to not be working properly?
- [x] add missing (non-multi) methods


src/nwc/NWCWalletService.ts
```diff
@@ -0,0 +1,294 @@
...
+    }
+    this.walletPubkey = getPublicKey(hexToBytes(this.walletSecret));
+  }
+}
+
+export class NWCWalletService {
+  relay: Relay;
+  relayUrl: string;
+
+  constructor(options: NewNWCWalletServiceOptions) {
```
bumi, line 52: why does the wallet service not also have the keypair?

bumi (commented):
that looks like a nice and helpful wrapper! 

do I read it correct that for each user/wallet I would need to call the subscribe function?

src/nwc/NWCWalletService.ts
rolznz: > do I read it correct that for each user/wallet I would need to call the subscribe function?

Yes, this is the reason. So that each connection can have unique keypairs, like we do in Alby Hub now

rolznz:
Merging for now and will add documentation in a follow-up PR to not cause conflicts with https://github.com/getAlby/js-sdk/pull/344

PR #356, opened 2025-04-07
feat: allow comment and payerdata in LNClient pay method

rolznz:
I'm not sure if it is clear enough that it can be used this way, but this seems the cleanest way to add it (maybe we can add an example?).


bumi:
Ah, args can contain the metadata. sounds good to me. let's use it in the example and readme and it's perfect.

PR #357, opened 2025-04-09
feat: add lnclient receive timeout for ReceiveInvoice class

rolznz:
Based on the updated comments, some things can be simplified/improved with the `onTimeout` method.

`ReceiveInvoice` has changed so it does not have async methods, so that they can be chained, and also the `async onPaid` function was a bit strange because all it returned was a ~~timeout~~ unsubscribe method (not once the invoice was paid).

### DIFF
--- a/src/lnclient/ReceiveInvoice.ts
+++ b/src/lnclient/ReceiveInvoice.ts
@@ -1,10 +1,17 @@
 import { Invoice } from "@getalby/lightning-tools";
 import { Nip47Notification, Nip47Transaction, NWCClient } from "../nwc";
 
+/**
+ * A lightning invoice to be received by your wallet, along with utility functions,
+ * such as checking if the invoice was paid and acting upon it.
+ */
 export class ReceiveInvoice {
   readonly transaction: Nip47Transaction;
   readonly invoice: Invoice;
   private _nwcClient: NWCClient;
+  private _unsubscribeFunc?: () => void;
+  private _timeoutFunc?: () => void;
+  private _timeoutId?: number | NodeJS.Timeout;
 
   constructor(nwcClient: NWCClient, transaction: Nip47Transaction) {
     this.transaction = transaction;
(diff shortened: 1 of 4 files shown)


src/lnclient/ReceiveInvoice.ts
```diff
@@ -1,64 +1,135 @@
...
+  onPaid(
+    callback: (receivedPayment: Nip47Transaction) => void,
+  ): ReceiveInvoice {
+    (async () => {
+      let supportsNotifications;
+      try {
+        // TODO: is there a better way than calling getInfo here?
+        const info = await this._nwcClient.getInfo();
+        supportsNotifications =
+          info.notifications?.includes("payment_received");
```
bumi, line 37: we can't do this in the constructor, or when the client connects?

bumi (approved):
looks like a great API. feels good. 
I don't fully grasp all of the code though. but love it

src/lnclient/ReceiveInvoice.ts
rolznz: This would be better, I don't like that we do it for every invoice like this

rolznz: I'm just wondering if it's right to put it in the `NWCClient` that we then slow down the connection to do this `get_info` call and not necessarily all clients would want this to happen by default. (and for apps that don't know about it will do another `get_info` call which would be redundant?)

bumi: this is the problem that we want to take away a task from the developer but at the same time we don't want that because we don't know what the developer actually does. 

in the end you have to pick one. imo the nwc client knowing the info of the nwc wallet makes sense - it is done only once in the lifecycle of the client. 
if we would want to make it "optional" it means more configuration and a more complex library.

but I am also OK with doing it each time in the onPaid callback. it has to wait for the payment from the user anyway

rolznz: Thanks for your thoughts, I agree it's not good to make it optional. Maybe it's ok as is for now and we can improve this as a follow-up since it's only internal details and the API shouldn't change?

I'll merge this PR for now and keep thinking about it / we wait for usage and actual issues in how it is implemented now.

PR #375, opened 2025-04-29
Migrate example JS files to TypeScript (Fixes #362)  

### LINKED ISSUE #362: Convert examples to Typescript
As JS the examples are more difficult to maintain.

Dunsin-cyber:
**Description:**  
This PR converts all JavaScript example files to TypeScript, including:  
- `examples/lnclient`  
- `examples/nwc`  
- `examples/oauth`  

While migrating, I found type errors or inconsistencies in a few files. Instead of mixing fixes with the migration, I’ve **commented them out** for now to be addressed separately in a follow-up issue. The affected files are:  
- `client/nwa-accept.ts`  
- `client/get-wallet-service-supported-methods.ts`  
- `wallet-service/example.ts`  
- `oauth/invoices`  

This change improves type safety and maintainability while keeping the examples functional. All tests pass for the converted files.  

**Next Steps:**  
A new issue should be opened to properly fix the commented-out files.


rolznz:
Hi @Dunsin-cyber , unfortunately this is very hard to review and has some strange things in here (importing from JS, TODOs, commented out code leaving previously working examples broken, files deleted and re-added rather than properly re-named).

It will take more time to review this properly than to do it from scratch. I wonder if we should do one file at a time per pull request, (and only do one pull request at a time).

Dunsin-cyber:
Got it, that works too. I'll create a separate PR for each file change.

Regarding the TODOs and commented-out code, since the previous examples were in JavaScript, I encountered issues after converting them to TypeScript. Some of the functions had incorrect parameters passed into them. While this can easily be fixed by passing the correct arguments, I didn't want to step outside the scope of the PR, which is why I added TODOs and commented them out, as they were causing errors.

Also, you mentioned importing JavaScript files. Could you clarify that? The only file I imported was the generated type from the `dist` folder so I can work on it in the next series of PRs.

rolznz:
Thanks. If there are issues in the examples it would be great to fix them at the same time - this is exactly the reason why we want to make this change in the first place so that these issues become clear.

> Also, you mentioned importing JavaScript files. Could you clarify that? The only file I imported was the generated type from the dist folder for type definitions, so I can work on it in the next series of PRs.

I mean this:
```
// Import runtime JS and types separately
import { LN, USD } from "../../dist/index.module.js";
import type { LNClient } from "../../dist/index";
```

Dunsin-cyber:
thanks for the clarification ,  I understand now. To confirm:
 
about this, 
> I mean this:
> 
> ```
> // Import runtime JS and types separately
> import { LN, USD } from "../../dist/index.module.js";
> import type { LNClient } from "../../dist/index";
> ```

I should use the proper package interface:

`import { LN, SATS, USD, type LNClient } from "@getalby/sdk";`


since 

```typescript
  "exports": {
    "require": "./dist/index.cjs",
    "types": "./dist/index.d.ts",
    "default": "./dist/index.modern.js"
  },
``` 

  
  is in the package.json already

rolznz:
@Dunsin-cyber this would not use the locally built version of the JS SDK though.

Dunsin-cyber:
yes.. do you have any suggestions or advice on how I can approach testing this in a different way without importing the built js file ?

PR #383, opened 2025-05-08
Rewrite pay_ln_address.js to TypeScript

Dunsin-cyber:
This pull request is based on my earlier discussion with @rolznz regarding rewriting each example file in TypeScript and submitting them as individual PRs. The goal is to:

* **Fix:** https://github.com/getAlby/js-sdk/issues/362
* **Related to:** #375

---

### Changes in This PR:

* Renamed `pay_ln_address.js` → `pay_ln_address.ts`
* Rewrote the file in TypeScript, using:

  * Type-safe imports for runtime (`index.module.js`)
  * Separate type import for `LNClient` from `dist/index.d.ts`
* Preserved existing functionality while improving readability and type safety

---

### Notes:

> Since the original `.js` file was fully rewritten as `.ts`, Git may detect this as a delete + add. Let me know if you'd prefer a two-step commit (`git mv` → then content edit) in future PRs for cleaner diff tracking.

---

@rolznz — starting with this file as the first step. Let me know if the structure, import style, and formatting align with your expectations. I’d be happy to adjust anything as needed so we can use this as the reference approach for the remaining conversions.

Looking forward to your feedback!


examples/lnclient/pay_ln_address.ts
```diff
@@ -0,0 +1,34 @@
+import "websocket-polyfill"; // required in node.js
+
+import * as readline from "node:readline/promises";
+import { stdin as input, stdout as output } from "node:process";
+
+// ⚠️ Using locally built SDK for testing — run `pnpm build` before running this example.
+import { LN, USD } from "../../dist/index.module.js";
```
rolznz, line 7: We want to avoid this as it defeats the whole purpose of moving to typescript. We either need to import the actual typescript files or import from `@getalby/sdk` and add a package.json to the examples folder.

Dunsin-cyber:
Alright.
Thanks for the feedback!

PR #387, opened 2025-05-09
chore: setup TypeScript support and package config in examples folder

Dunsin-cyber:
This PR sets up a standalone environment in the `examples/` folder to support running TypeScript-based usage examples for @getalby/sdk.

### Changes:
- Added `package.json` to examples folder with local file dependency to SDK.
- Configured `ts-node` for running TS files directly.
- Created basic `README.md` with setup and usage instructions.

This setup will be used as the foundation for rewriting and running individual example files in TypeScript, as discussed with @rolznz in issue #362 and PR #383.


rolznz:
Hi @Dunsin-cyber could you include one updated example file in the PR so I can test it works?

Dunsin-cyber:
alright, on it.

Dunsin-cyber:
Hi @rolznz , I've made the following changes:

Renamed the (pay_ln_address.js) example file properly (preserving git history).

Added one updated example (pay_ln_address.ts) to the PR.

Included a Prerequisites section in the README.md to guide usage with the local SDK build.

examples/README.md
```diff
@@ -0,0 +1,57 @@
...
+yarn install
+```
+
+
+## ▶️ Running an Example
+
+Use `ts-node` to run a specific TypeScript file. For example:
+
+```bash
+yarn ts-node ./lnclient/pay_ln_address.ts
```
rolznz, line 38: I get 
```
TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /home/roland/dev/alby/js-sdk/examples/lnclient/pay_ln_address.ts
```
When running on both node 20 and node 22. Did you run the example?

Dunsin-cyber:
Hi @rolznz, thanks for pointing that out.

It worked on my end initially,  I likely had a global ts-node configuration that handled the setup automatically. I wasn't able to reproduce the error at first, but I managed to replicate it when testing on Windows.

I've now updated the example to use tsx instead of ts-node to avoid extra set-up that comes with ts-node, since tsx works out of the box with minimal configuration. I’ve tested the fix on both environments, and it should now run without issues.

I would love if you could give it another try and let me know if you run into anything else!

rolznz:
@Dunsin-cyber I've pushed some improvements, could you please review the changes and let me know what you think?

Dunsin-cyber:
> @Dunsin-cyber it kind of works when I add a .yarnrc.yml file with `nodeLinker: node-modules` - without it I got an out of memory error on Linux.
> 
> But it's not an optimal flow right now. Every time I run `yarn build` in the JS SDK, I also have to re-run `yarn install` in the examples folder, which is quite slow as it copies 600+MB of files from the parent directory into the current one.
> 
> Changing the import in the example to `import { LN, USD } from "../../src/lnclient/index.ts";` works but not `import { LN, USD } from "../../src/index.ts";` which would be more ideal as that can easily be swapped for the real import ("@getalby/sdk"). It seems to have some issues compiling. Any chance you could have a look and see if you can get that to work?
> 
> Then we don't need to add the JS SDK as a dependency in the examples folder and imports work almost exactly as they do now, but typescript instead of javascript.
> 
> Note: we will need to add `"websocket-polyfill": "^0.0.3"` as a dependency in the examples package.json when we remove the JS SDK dependency.

Dunsin-cyber:
> Changing the import in the example to `import { LN, USD } from "../../src/lnclient/index.ts";` works but not `import { LN, USD } from "../../src/index.ts";` which would be more ideal as that can easily be swapped for the real import ("@getalby/sdk"). It seems to have some issues compiling. Any chance you could have a look and see if you can get that to work?
> 
> Then we don't need to add the JS SDK as a dependency in the examples folder and imports work almost exactly as they do now, but typescript instead of javascript.

I am still looking into this, will get back to you shortly

Dunsin-cyber:
> Changing the import in the example to `import { LN, USD } from "../../src/lnclient/index.ts";` works but not `import { LN, USD } from "../../src/index.ts";` which would be more ideal as that can easily be swapped for the real import ("@getalby/sdk"). It seems to have some issues compiling. Any chance you could have a look and see if you can get that to work?

@rolznz  so i was able to replicate the error you were referring to when i run ` yarn tsx ./lnclient/pay_ln_address.ts` with `import { LN, USD } from "../../src/index";` here; 
```bash
/home/dunsin/gh/fork/js-sdk/src/oauth/types.ts:2
import { RequestOptions } from "./request";
         ^

SyntaxError: The requested module './request' does not provide an export named 'RequestOptions'
```
i was able to fix it by  using `import type { RequestOptions } from "./request";` instead because TypeScript assumes we're importing a real JavaScript value, something that exists at runtime. But since `RequestOptions` is just a TypeScript type (which disappears after compilation), this can cause a runtime error like: `ERR_MODULE_NOT_FOUND`

Dunsin-cyber:
alrigh, I've reverted the changes, @rolznz. ⚡

if i may ask, what about this redundancy you mentioned with portal

> But it's not an optimal flow right now. Every time I run `yarn build` in the JS SDK, I also have to re-run `yarn install` in the examples folder, which is quite slow as it copies 600+MB of files from the parent directory into the current one.

rolznz:
@Dunsin-cyber with portal it doesn't copy - it just creates a symlink. Therefore you only need to run yarn install once and the directory will always be updated when you build the JS SDK again

Dunsin-cyber:
oh... yes. that's true!

PR #393, opened 2025-05-21
chore: add recipient_data to tx metadata

im-adithya:
Only adds identifier as description is not widely used and is usually the same message everytime (ex: "Sats for Adithya Vardhan")

### DIFF
--- a/src/nwc/types.ts
+++ b/src/nwc/types.ts
@@ -161,6 +161,16 @@ export type Nip47Transaction = {
 };
 
 export type Nip47TransactionMetadata = {
+  success_action?:
+    | {
+        tag: "message";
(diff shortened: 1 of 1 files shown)


src/nwc/types.ts
```diff
@@ -171,6 +171,9 @@ export type Nip47TransactionMetadata = {
     pubkey: string;
     tags: string[][];
   }; // NIP-57
+  recipient_data?: {
+    identifier: string;
```
rolznz, line 175: can you make this optional too?

im-adithya: Done!

rolznz (approved):

PR #405, opened 2025-05-30
Refactor:  spliiter.js to ts to show split payment example

Dunsin-cyber:
this rewrites/refactors  `spliiter.js` found in `examples/lnclient` in typescript as suggested here https://github.com/getAlby/js-sdk/pull/375#issuecomment-2840728478


rolznz:
Did you do a self review? and in the future can you rename the file and commit it first before making changes? otherwise we get a deleted file and new file added in our history.

Dunsin-cyber:
Yes, I tested it and everything worked as expected.

actually, that was what I did.  rename the file first and committed that separately before making any actual changes. GitHub reflects this in the commit history,  you’ll see two separate commits showing that process.

However, I found out that when the content of the renamed file changes significantly, Git sometimes treats it as a file deletion and creation, rather than an update to a renamed file. So that might be why it appears that way now.

examples/lnclient/splitter.ts
```diff
@@ -4,61 +4,101 @@ import qrcode from "qrcode-terminal";
...
-    const satsToForward = Math.floor(
-      (satsReceived * forwardPercentage) / 100 / recipients.length,
-    );
-    console.info(
-      `Received ${satsReceived} sats! Forwarding ${satsToForward} to the ${recipients.join(", ")}`,
-    );
+      // iterate over all recipients and pay them the amount
+      await Promise.all(
+        recipients.map(async (r) => {
+          const response = await client.pay(r, SATS(satsToForward), {
```
rolznz, line 79: could you keep the description as it was before

examples/lnclient/splitter.ts
```diff
@@ -4,61 +4,101 @@ import qrcode from "qrcode-terminal";
...
+    description: "[TEST] Split Payment Example",
+  });
+
+  // prompt the user to pay the invoice
+  qrcode.generate(request.invoice.paymentRequest, { small: true });
+  console.info(request.invoice.paymentRequest);
+  console.info("Please pay the above invoice within 60 seconds.");
+  console.info("Waiting for payment...");
+
+  //the feilds here are optional, but can be used to store metadata about the payment
```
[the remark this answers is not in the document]
Dunsin-cyber: alright

examples/lnclient/splitter.ts
Dunsin-cyber: alright
i just pushed the fix.

Dunsin-cyber:
oh.. alright.
for subsequent PRs, what should we do about the file rename appearing as file being deleted?

basically what i do is 

- git mv file.js file.ts

- then i commit the file rename 

- make my changes and make final commit

PR #421, opened 2025-06-23
feat(examples/nwc) ; rewrite get-info in typescript

Dunsin-cyber:
this rewrites `get-info.js` found in `examples/nwc` in typescript as suggested here #375


rolznz (approved):
utACK

PR #425, opened 2025-07-02
feat(examples/nwc) ; rewrite keysend in typescript 

Dunsin-cyber:
this rewrites `keysend.js` found in examples/nwc in typescript as suggested here https://github.com/getAlby/js-sdk/pull/375


rolznz (approved):
utACK

PR #427, opened 2025-07-02
feat(examples/nwc) : rewrite list-transactions in typescript 

Dunsin-cyber:
this rewrites `list-transactions.js` found in examples/nwc in typescript as suggested here https://github.com/getAlby/js-sdk/pull/375


rolznz (approved):
utACK

PR #443, opened 2025-08-01
feat: add subpath imports and fix bundling issues

im-adithya:
## ⚠️ Breaking Change

This fixes the long pending bundling issues using rollup.

We now export both esm and cjs files, and also types in such a way that devs can do either:

```js
import { CreateSwapResponse } from "@getalby/sdk";
// or
import { CreateSwapResponse } from "@getalby/sdk/oauth";
// actually it's best to do it like this so compiler understands that we only need the typing
import type { CreateSwapResponse } from "@getalby/sdk/oauth";
```

And this removes export as nwc, oauth etc. So now:
```js
import { nwc } from "@getalby/sdk";
const nwcClient = new nwc.NWCClient({
  nostrWalletConnectUrl: loadNWCUrl(),
});
```

will not work, we should instead do:
```js
import { NWCClient } from "@getalby/sdk/nwc";
// or
import { NWCClient } from "@getalby/sdk";
const nwcClient = new NWCClient({
  nostrWalletConnectUrl: loadNWCUrl(),
});
```

## TODOs
- [x] Fix examples
- [x] Test `lightning-sdk.bundle.js`
- [x] Test esm
- [x] Test cjs
- [x] Test types
- [x] Test subpath imports
- [x] Test normal imports

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,25 +2,46 @@
   "name": "@getalby/sdk",
   "version": "5.1.1",
   "description": "The SDK to integrate with Nostr Wallet Connect and the Alby API",
+  "type": "module",
   "repository": "https://github.com/getAlby/js-sdk.git",
   "bugs": "https://github.com/getAlby/js-sdk/issues",
   "funding": {
     "type": "lightning",
     "url": "lightning:hello@getalby.com"
   },
-  "type": "module",
-  "source": "src/index.ts",
-  "main": "./dist/index.umd.js",
-  "module": "./dist/index.module.js",
-  "unpkg": "./dist/index.umd.js",
-  "types": "./dist/index.d.ts",
   "files": [
-    "dist/**/*"
+    "dist"
   ],
+  "sideEffects": false,
+  "module": "./dist/esm/index.js",
+  "main": "./dist/cjs/index.cjs",
+  "types": "./dist/types/index.d.ts",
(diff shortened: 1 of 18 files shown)


rolznz:
@im-adithya it only breaks tree shaking it if those imports are actually used, right? so we can update our examples and documentation to not use them, but I wonder if we should also keep them in. I am not sure if it's necessary to do a breaking change here if we don't have to

rolznz:
@im-adithya update: from our call we decided to just do the breaking change

package.json
```diff
@@ -2,25 +2,46 @@
...
   "bugs": "https://github.com/getAlby/js-sdk/issues",
   "funding": {
     "type": "lightning",
     "url": "lightning:hello@getalby.com"
   },
-  "type": "module",
-  "source": "src/index.ts",
-  "main": "./dist/index.umd.js",
-  "module": "./dist/index.module.js",
-  "unpkg": "./dist/index.umd.js",
```
rolznz, line 15: e.g. also this one

rollup.config.js
```diff
@@ -0,0 +1,66 @@
+import resolve from "@rollup/plugin-node-resolve";
```
rolznz, line 1: Can you link to docs for how this file was created?

PR #465, opened 2025-08-13
Convert get-balance to Typescript

Dunsin-cyber:
this rewrites `get-balance.js` found in examples/nwc/client to typescript as suggested here https://github.com/getAlby/js-sdk/pull/375#issuecomment-2840728478


rolznz (approved):
utACK

PR #467, opened 2025-08-13
Convert get-info to Typescript

Dunsin-cyber:
this rewrites `get-info.js` found in examples/nwc/client to typescript as suggested here https://github.com/getAlby/js-sdk/pull/375#issuecomment-2840728478


rolznz:
@Dunsin-cyber if you see a lot of examples that only require minor edits like this one, feel free to put multiple in the same PR. Thanks for updating all of these! :muscle:

Dunsin-cyber:
> @Dunsin-cyber if you see a lot of examples that only require minor edits like this one, feel free to put multiple in the same PR. Thanks for updating all of these! 💪

alright, will do. 
my pleasure 😊

PR #471, opened 2025-09-01
Rewrite Examples in examples/oauth in Typescript

Dunsin-cyber:
this rewrites all examples  found in `examples/oauth` to typescript as suggested here https://github.com/getAlby/js-sdk/pull/467#issuecomment-3184351715 and  https://github.com/getAlby/js-sdk/pull/375#issuecomment-2840728478

### DIFF
--- a/examples/oauth/delete-webhook.ts
+++ b/examples/oauth/delete-webhook.ts
@@ -1,0 +1,49 @@
+import * as readline from "node:readline/promises";
+import { stdin as input, stdout as output } from "node:process";
+
+import { Client, OAuth2User } from "@getalby/sdk/oauth";
+
+const rl = readline.createInterface({ input, output });
+
+if (!process.env.CLIENT_ID || !process.env.CLIENT_SECRET) {
+  throw new Error("Please set CLIENT_ID and CLIENT_SECRET");
+}
+
+
+const authClient = new OAuth2User({
+  client_id: process.env.CLIENT_ID,
+  client_secret: process.env.CLIENT_SECRET,
+  callback: "http://localhost:8080/callback",
+  user_agent:"AlbySDK-Example/0.1 (delete_webhook-demo)",
+  scopes: [
+    "invoices:read",
+    "account:read",
+    "balance:read",
+    "invoices:create",
+    "invoices:read",
+    "payments:send",
+  ],
+  token: {
+    access_token: undefined,
+    refresh_token: undefined,
+    expires_at: undefined,
+  }, // initialize with existing token
+});
+
+console.log(`Open the following URL and authenticate the app:`);
+console.log(await authClient.generateAuthURL());
+console.log("----\n");
+
+const code = await rl.question("Code: (localhost:8080?code=[THIS CODE]: ");
+
+const webhookId = await rl.question("Enter the webhook ID to delete: ");
+rl.close();
+
+await authClient.requestAccessToken(code);
+console.log(authClient.token);
+const client = new Client(authClient);
+
+
+// Delete a webhook
+const deleteResult  = await client.deleteWebhookEndpoint(webhookId)
+console.log("webhook deleted", JSON.stringify(deleteResult));


Dunsin-cyber:
> Hi @Dunsin-cyber thanks! happy new month to you too :-)
> 
> By the way you added an image with some OAuth codes in it. I removed that image (to avoid someone being able to access your Alby Account.)
> 
> I don't think the `q` parameter in `GetInvoicesRequestParams` is handled properly. In the docs we pass as `q[created_at_lt]` but how we encode it when doing the request seems to be with this `buildQueryString` function which doesn't handle the embedded values properly (it makes something like `q=[object Object]` which will not work.
> 
> Would you like to try doing a PR to fix it?
> 
> Regarding keysend, I can't see any reason why it would return an empty response. You're Alby Account is connected to your Alby Hub right? I don't think we support sending the custom record `34349334`

Ohh the URL, thanks, thank you.
alright on it. i'd make a separate PR to address that.

i'd double-check on the `keysend` issue again

Dunsin-cyber:
> Hi @Dunsin-cyber thanks! happy new month to you too :-)
> 
> By the way you added an image with some OAuth codes in it. I removed that image (to avoid someone being able to access your Alby Account.)
> 
> I don't think the `q` parameter in `GetInvoicesRequestParams` is handled properly. In the docs we pass as `q[created_at_lt]` but how we encode it when doing the request seems to be with this `buildQueryString` function which doesn't handle the embedded values properly (it makes something like `q=[object Object]` which will not work.
> 
> Would you like to try doing a PR to fix it?
> 
> Regarding keysend, I can't see any reason why it would return an empty response. You're Alby Account is connected to your Alby Hub right? I don't think we support sending the custom record `34349334`

so `keysends.ts` and `boostagram.ts` (because it uses keysend under the hood) are failing.

i have used the following node IDs, with `customRecords: { 34349334: "I love amboss" },` and no custom records
- 03006fcf3312dae8d068ea297f58e2bd00ec1ffe214b793eda46966b6294a53ce6
- 03390c79737f525a07e229a4c50cb6259d687859ad2622707565bd3881e4355736
- 02947ea84b359c2e902c10e173aa209a36c2f92a6143c73170eb72b2077c592187

i still get an empty object, and yes my Alby Hub is connected.
the only time i get a thrown error (after it returns same empty object) is when i do only one keysend and not multiple keysends
i.e when i hit this endpoint `  endpoint = "/payments/keysend";` and not this `  endpoint = "/payments/keysend/multi";`
```bash
{}
/home/dunsin/gh/fork/js-sdk/src/oauth/request.ts:79
    throw new AlbyResponseError(
          ^


AlbyResponseError: 400 Bad Request: NodeError: PaymentSendingFailed: Failed to send the given payment.
    at request (/home/dunsin/gh/fork/js-sdk/src/oauth/request.ts:79:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async rest (/home/dunsin/gh/fork/js-sdk/src/oauth/request.ts:91:20) {
  status: 400,
  statusText: 'Bad Request',
  headers: Headers {},
  error: {
    error: true,
    message: 'NodeError: PaymentSendingFailed: Failed to send the given payment.'
  }
}

Node.js v23.11.0
```

failed txs

rolznz:
@Dunsin-cyber thanks for checking, I'll add to my TODO list to investigate the keysend issue

rolznz:
@Dunsin-cyber sorry, not yet. The backend looks OK but I need to reproduce the error. How do you run the OAuth examples? I need to run an OAuth server locally? I wonder if we could add support for optionally using an access token instead of a CLIENT_ID and CLIENT_SECRET? from https://getalby.com/developer/access_tokens/new

Maybe we can also make a README file in the examples/oauth folder too so users know how to more easily try the OAuth examples with instructions for both methods? what do you think?

rolznz:
@Dunsin-cyber when I followed your instructions I get:

> Error: Please set CLIENT_ID and CLIENT_SECRET

So it would be good to include instructions in the example on how to get these. Another PR sounds good.

examples/oauth/webhooks.ts
```diff
@@ -37,12 +42,13 @@ console.log(authClient.token);
...
+  filter_types: ["invoice.incoming.settled", "invoice.outgoing.settled"],
 });
+console.log("webhook created", JSON.stringify(webhook));
 
-// Delete a webhook
-// response = await alby.deleteWebhookEndpoint('ep_...').then(console.log)
 
-console.log(JSON.stringify(response));
+// Delete a webhook
+const deleteResult  = await client.deleteWebhookEndpoint(webhook.id)
```
rolznz, line 53: can we move deletion to another example (delete_webhook) and rename this one to create_webhook

And use a real url for the webhook - maybe the user can be prompted and they can use a url from https://webhook.site/

And we should print out the ID in the create_webhook example so that the user can pass it to the delete_webhook example.

Then developers can actually test the webhook works.

Dunsin-cyber: for the naming convention, most of the examples  were in dash case, i.e `make-invoice`, `sign-message`.
 
Should we still follow that or go with this snake case `delete_webhook`

rolznz: Ah ok, dash case is good

Dunsin-cyber: Done

Dunsin-cyber:
hey @rolznz, i’ve made the adjustments.

rolznz (approved):
tACK, nice work!

Dunsin-cyber:
> tACK, nice work!

thanks😊

PR #480, opened 2025-09-24
feat(oauth): introduce helper for auth flow + test in invoices.ts

### LINKED ISSUE #478: Enhance OAuth examples to include Access Token support
I wonder if we could add support for optionally using an access token instead of a CLIENT_ID and CLIENT_SECRET? from https://getalby.com/developer/access_tokens/new

Maybe we can also make a README file in the examples/oauth folder too so users know how to more easily try the OAuth examples with instructions for both methods? what do you think?

_Originally posted by @rolznz in https://github.com/getAlby/js-sdk/issues/471#issuecomment-3291453795_

Dunsin-cyber:
fixes #478
@rolznz
i created a helper function to handle the auth flow and tested it in `invoices.ts`.

if this approach looks good to you, i’ll go ahead and add the helper function to the rest of the oauth examples.

you can follow the readme to test it out.

### DIFF
--- a/examples/oauth/helper.ts
+++ b/examples/oauth/helper.ts
@@ -2,12 +2,12 @@ import * as readline from "node:readline/promises";
 import { stdin as input, stdout as output } from "node:process";
 import dotenv from "dotenv";
 dotenv.config();
-import { OAuth2User } from "@getalby/sdk/oauth";
-
-const rl = readline.createInterface({ input, output });
+import { Client, OAuth2User } from "@getalby/sdk/oauth";
 
 async function getAuthClient(user_agent: string) {
   if (process.env.CLIENT_ID && process.env.CLIENT_SECRET) {
+    const rl = readline.createInterface({ input, output });
+
     console.log("🔑 Using OAuth2 flow with CLIENT_ID and CLIENT_SECRET...");
     const authClient = new OAuth2User({
       client_id: process.env.CLIENT_ID,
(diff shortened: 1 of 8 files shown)


examples/oauth/helper.ts
```diff
@@ -0,0 +1,48 @@
...
+    const code = await rl.question("Code: (localhost:8080?code=[THIS CODE]: ");
+    rl.close();
+
+    await authClient.requestAccessToken(code);
+    console.log(authClient.token);
+
+    return authClient;
+  } else if (process.env.ACCESS_TOKEN) {
+    console.log("🔒 Using direct Access Token from environment...");
+    rl.close();
```
rolznz, line 38: I guess `rl` should not be used here and only initialized when using the CLIENT_ID and CLIENT_SECRET

examples/oauth/invoices.ts
```diff
@@ -1,43 +1,16 @@
...
-    access_token: undefined,
-    refresh_token: undefined,
-    expires_at: undefined,
-  }, // initialize with existing token
-});
-
-console.log(`Open the following URL and authenticate the app:`);
-console.log(await authClient.generateAuthURL());
-console.log("----\n");
+const client = new Client(auth);
```
rolznz, line 8: maybe we could directly return the `new Client(auth)` inside the helper?

examples/oauth/helper.ts
Dunsin-cyber: yea, that's true. removed

Dunsin-cyber:
> @Dunsin-cyber nice work! just minor comments.

thanks 💪🏾

PR #530, opened 2026-02-13
fix: NWCClient.close() hanging Jest tests

### LINKED ISSUE #525: Bug: sockets are not properly closed
When writing jest tests with AI (alby agent skill) the jest tests don't exit by themselves, and the AI hacks it by adding `jest --forceExit`

```bash
npx jest --detectOpenHandles 2>&1 | tail -30
```

Review the below - might need to just update nostr-tools or there needs to be a fix there or we call close incorrectly on our side.

What AI says:
> The open handle is a TCPWRAP — a lingering WebSocket/TCP connection from the nostr-tools relay pool inside   
  NWCClient.

```txt
The bug: NWCClient.close() is synchronous and returns before WebSocket connections are fully terminated.     
                                                                                                               
  Call chain:                                                                                                  
                                                                                                               
  1. NWCClient.close() calls this.pool.close(this.relayUrls) — src/nwc/NWCClient.ts                            
  2. SimplePool.close() calls relay.close() on each relay, then deletes it from the map                        
  3. AbstractRelay.close() (in nostr-tools) calls this.ws?.close() — which only initiates the WebSocket close  
  handshake. The underlying TCP socket stays alive until the server responds with a close frame.               
                                                                                                               
  Because NWCClient.close() is synchronous and ws.close() is non-blocking, the method returns immediately while
   the TCP socket is still open mid-handshake. This is the dangling TCPWRAP handle that prevents Node.js/Jest  
  from exiting.                                                                                                
                                                                                                               
  The fix in @getalby/sdk: Make NWCClient.close() async so it can wait for the WebSocket connections to        
  actually close. Before calling this.pool.close() (which clears the relay map), capture references to each    
  relay's ws, then await their close events:                                                                   
                                                                                                               
  // In NWCClient (src/nwc/NWCClient.ts)                                                                       
  async close(): Promise<void> {                                                                               
    // Grab WebSocket refs before pool.close() clears the relay map                                            
    const closePromises: Promise<void>[] = [];                                                                 
    for (const url of this.relayUrls) {                                                                        
      const relay = this.pool.relays.get(normalizeURL(url));                                                   
      if (relay?.ws && relay.ws.readyState !== WebSocket.CLOSED) {                                             
        closePromises.push(                                                                                    
          new Promise<void>((resolve) => {                                                                     
            relay.ws!.addEventListener("close", () => resolve(), { once: true });                              
          })                                                                                                   
        );                                                                                                     
      }                                                                                                        
    }                                                                                                          
                                                                                                               
    this.pool.close(this.relayUrls);                                                                           
                                                                                                               
    await Promise.all(closePromises);                                                                          
  }                                                                                                            
                                                                                                               
  This is backward-compatible — callers that don't await get the same fire-and-forget behavior as before, while
   callers that do await (like test teardown) can ensure clean shutdown.
```

AkarshSahlot:
Closes #525

The close() method in NWCClient was synchronous, which meant it didn't wait for WebSocket connections to actually close before returning. This caused Jest tests to hang because of open handles.

I made close() async so it waits for the underlying WebSocket capabilities to finish closing. I also updated the dependencies and config to support the latest nostr-tools exports.

Changes:
- Made NWCClient.close() async
- Updated imports to use specific paths from nostr-tools
- Set moduleResolution to bundler in tsconfig
- Updated jest.config.ts to export default

## Summary by CodeRabbit

* **Refactor**
  * Improved resource cleanup logic for better connection management.

* **Chores**
  * Updated build and test configurations for ESModule support.
  * Modernized internal module imports and resolution paths.

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,5 +1,8 @@
 /** @type {import('ts-jest').JestConfigWithTsJest} */
-module.exports = {
-  preset: 'ts-jest',
-  testEnvironment: 'node',
+export default {
+  preset: "ts-jest",
+  testEnvironment: "node",
+  transform: {
+    "^.+\\.tsx?$": ["ts-jest", { useESM: true }],
+  },
 };
--- a/src/nwc/NWAClient.ts
+++ b/src/nwc/NWAClient.ts
@@ -7,7 +7,7 @@ import {
   Nip47NotificationType,
 } from "./types";
 import { NWCClient } from "./NWCClient";
-import { SubCloser } from "nostr-tools/lib/types/abstract-pool";
+import { SubCloser } from "nostr-tools/abstract-pool";
 
 export type NWAOptions = {
   relayUrls: string[];
--- a/src/nwc/NWCClient.ts
+++ b/src/nwc/NWCClient.ts
@@ -10,6 +10,7 @@ import {
   EventTemplate,
   SimplePool,
 } from "nostr-tools";
+import { normalizeURL } from "nostr-tools/utils";
 import { hexToBytes, bytesToHex } from "@noble/hashes/utils";
 import {
   Nip47EncryptionType,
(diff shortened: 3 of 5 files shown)


rolznz:
Closing, this is a hack and didn't solve the root issue

Issue #231, opened 2024-07-17
Feature request: Add debug mode option


dsbaars:
There are quite some console.log/info/error statements which can be useful for debugging, but also pollute the console.
It would be nice if NWCOptions could be extended with a "verbose" or "debug" option to control debug logging.

rolznz:
Completed by https://github.com/getAlby/js-sdk/pull/542

PR #541, opened 2026-04-07
fix: remove console.info statements that pollute stdout

reneaaron:
## Summary
- Remove all `console.info` calls across NWAClient, NWCClient, NWCWalletService, and OauthWeblnProvider
- Agents and tools consuming the SDK (e.g. via lightning-tools) were seeing log lines mixed into stdout, which could break parsing

## Test plan
- [ ] Verify SDK no longer logs to stdout during normal operation
- [ ] Run existing tests to confirm no regressions

## Summary by CodeRabbit

* **Chores**
  * Updated logging configuration to adjust console output levels across the application for improved diagnostic output.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,7 @@
       }
     ],
     "@typescript-eslint/no-unused-vars": ["warn", { "args": "none" }],
-    "no-console": ["error", { "allow": ["info", "warn", "error"] }],
+    "no-console": ["error", { "allow": ["debug", "warn", "error"] }],
     "no-constant-binary-expression": "error"
   },
   "globals": {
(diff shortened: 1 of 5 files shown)


rolznz:
Linking to https://github.com/getAlby/js-sdk/issues/231 - rather than removing them completely, I like the idea of a verbose mode (and do we need to check debug, etc level logs too?). What do you think?

reneaaron:
This is a first step to address stdout pollution — switching `console.info` to `console.debug` so the SDK stays silent on stdout by default.

As a follow-up, we should plan how proper log-level support could work for this library (e.g. a `debug` option, the `debug` npm package, or a pluggable logger interface) so that consumers can opt into debug/info/warn output when needed.

reneaaron:
Closing in favor of #542

PR #538, opened 2026-03-26
feat(e2e): add NWC contract tests for get_*, hold, multi, and notifications

DSanich:
**Summary**

- Extends the NWC API **contract-style e2e** suite beyond payment-only flows: `get_info`, `get_budget`, faucet `get_balance` (approximate balance vs exact faucet amount), `lookup_invoice` (by invoice string and by **`payment_hash` only**), `list_transactions` (outgoing after pay and **incoming** on the receiver), hold invoices (**make / settle / cancel** with two-wallet flows and realistic timing), **`multi_pay_invoice`** / **`multi_pay_keysend`**, and **`payment_received`** notifications (**subscribe before pay** so the relay subscription is active).
- Adds **`sign_message`** e2e: success when allowed, or **`NOT_IMPLEMENTED`** / **`RESTRICTED`** when the connection lacks scope—without assuming a single error code for “unsupported”.
- Tightens tests by removing misleading “any failure ⇒ NOT_IMPLEMENTED” patterns where they masked real bugs, and by modeling **cancel_hold_invoice** correctly (hold in-flight; safe handling of **`pay_invoice`** rejection / promise lifecycle after cancel).

**Tests**

- End-to-end coverage against the NWC faucet for the flows above, with capability checks, strict success paths where appropriate, and contract-safe outcomes where wallets differ (`NOT_IMPLEMENTED`, `RESTRICTED`, payer errors after hold cancel, etc.).

**SDK / client changes**

- None in this PR. Reliability fixes for **`subscribeNotifications`** (wait until the initial relay subscription is ready) and **`getWalletServiceInfo`** (retries when kind **13194** is missing transiently) live in a separate PR: branch **`fix/nwc-client-relay-subscription`**. For green CI on notification- and timing-sensitive flows, merge that client PR first (or temporarily base this branch on it), then rebase onto `master` after it lands.

PR #545, opened 2026-04-16
fix(nwc): stabilize relay subscription and wallet service info

DSanich:
## Summary

Stabilizes `NWCClient` behavior around relay subscriptions and discovery of the wallet service info event (NIP-47 kind **13194**). These changes were split out from the NWC contract e2e work so client behavior can be reviewed on its own.

## Motivation

1. **`subscribeNotifications`** previously returned the unsubscribe handle as soon as the subscription loop started. Callers could register handlers *after* the client had already subscribed, which made it easy to **miss notifications** that arrived immediately after the subscription became active.

2. **`getWalletServiceInfo`** could observe an **empty result** on the first `EOSE` (e.g. relay or timing quirks), and then throw even though a valid info event might appear shortly after.

## Changes

- **`subscribeNotifications`**: resolve a one-shot promise when the first successful relay subscription is established, and **`await`** it before returning the unsubscribe function to callers.
- **`getWalletServiceInfo`**: on empty results, **retry** up to a limited number of times with **incremental backoff** before failing with `no info event (kind 13194) returned from relay`.

## Scope

- `src/nwc/NWCClient.ts` only.


DSanich:
@rolznz Hello! This is a continuation of your message [here](https://github.com/getAlby/js-sdk/pull/538#issuecomment-4253626136). I've moved the changes related to NWCClient into a separate PR. So, we have two PRs. Please take a look:

- https://github.com/getAlby/js-sdk/pull/545
- https://github.com/getAlby/js-sdk/pull/538

PR #546, opened 2026-05-08
fix(nwc): add p tag to NIP-47 response events

ajaysehwal:
## Summary

Add the missing `p` tag to NIP-47 response events (`kind: 23195`).

Previously, wallet responses were published with only the request reference tag:

```ts
tags: [["e", event.id]]
```

This PR updates response events to include the client pubkey as well:

```ts
tags: [
  ["p", keypair.clientPubkey],
  ["e", event.id],
]
```

## Why

Some NIP-47 clients subscribe to wallet responses using `#p` filters:

```json
{
  "kinds": [23195],
  "#p": ["<client-pubkey>"]
}
```

Without the `p` tag, the relay does not forward the response event to those clients, even though the event is successfully published. This causes requests to eventually fail with a timeout.

Including both:

* `e` → request correlation
* `p` → recipient routing/filtering

improves interoperability with NIP-47 clients and aligns with the NIP-47 recommendation that response events SHOULD include a `p` tag.

## Summary by CodeRabbit

* **Bug Fixes**
  * Nostr Wallet Connect responses now include both the referenced request identifier and the client public key in response metadata, improving request correlation and tracking.

[Review Change Stack](https://app.coderabbit.ai/change-stack/getAlby/js-sdk/pull/546)

### DIFF
--- a/src/nwc/NWCWalletService.ts
+++ b/src/nwc/NWCWalletService.ts
@@ -196,7 +196,10 @@ export class NWCWalletService {
               const responseEventTemplate: EventTemplate = {
                 kind: 23195,
                 created_at: Math.floor(Date.now() / 1000),
-                tags: [["e", event.id]],
+                tags: [
+                  ["e", event.id],
+                  ["p", keypair.clientPubkey],
+                ],
                 content: await this.encrypt(
                   keypair,
                   JSON.stringify({


ajaysehwal:
@rolznz , could you please take sometime to review this PR ?

src/nwc/NWCClient.ts
```diff
@@ -984,6 +983,7 @@ export class NWCClient {
             kinds: [23195],
             authors: [this.walletPubkey],
             "#e": [event.id],
+            "#p": [this.publicKey],
```
rolznz: Hi, this increases the filter, not increases what is published. This might break compatibility with some apps. I'm not sure it's best to do this. @frnandu any thoughts?

src/nwc/NWCWalletService.ts
```diff
@@ -196,7 +196,10 @@ export class NWCWalletService {
               const responseEventTemplate: EventTemplate = {
                 kind: 23195,
                 created_at: Math.floor(Date.now() / 1000),
-                tags: [["e", event.id]],
+                tags: [
```
rolznz, line 199: This looks good. Do you use the NWCWalletService out of curiosity?

ajaysehwal: Yes, I’ve been working on the Nostr Wallet Connect service in the Zeus app, and we were seeing request timeout errors with several Nostr clients like Damus and Amethyst.
After debugging the issue, I believe the root cause is that some Nostr clients subscribe to NIP-47 response events (kind: 23195) using the p tag (the client’s public key). However, the current version of @getalby/sdk only includes the e tag in response events and does not include the p tag.
As a result, the wallet publishes the response successfully, but the response does not match the client’s subscription filter. The relay therefore never forwards the event to the client, and the client eventually reports a request timeout error.

src/nwc/NWCClient.ts
ajaysehwal: @rolznz  Good point — I think you're right.
Adding #p to the subscription filter makes the response matching more restrictive and could break compatibility with wallets that only include the e tag in their response events. I’ve reverted this change and kept the client-side subscription filter based on #e only.

rolznz (approved):
utACK

rolznz:
@ajaysehwal some other PRs need to be merged before the next release so this is available, I'll try to do a new release as soon as possible (it's been a while!)

PR #547, opened 2026-05-12
fix: remove websocket polyfill

### LINKED ISSUE #519: Remove websocket-polyfill
websocket-polyfill is only needed for NodeJS and only for < 22, and it incorrectly overwrites WebSocket even if it already exists on the global object.

It can remain as a small note in our README but should be removed from the rest of the project.

rolznz:
Closes https://github.com/getAlby/js-sdk/issues/519


rolznz:
E2E tests pass locally, seems like a temporary networking issue.

Issue #577, opened 2026-08-13
NWCWalletService.subscribe(): Unbounded replay of retained relay history


ajaysehwal:
### Summary

`NWCWalletService.subscribe()` builds its NIP-47 request filter with no `since`/`until` bound:

```js
const filter = {
  kinds: [23194],
  authors: [keypair.theirPubkey],
  "#p": [keypair.pubkey]
};
```

Because kind `23194` is in the ephemeral range, compliant relays are expected to drop it — but nothing enforces that, and a relay that
retains it (maliciously, or simply non-compliant) will redeliver the *entire* history of every request event on every fresh `REQ`. Since
consuming apps commonly tear down and recreate the subscription on every app restart, reconnect, or network change, this means a wallet
service built on this SDK can re-receive and re-process arbitrarily old requests every time it resubscribes.

The SDK's only replay defense is an in-memory `knownIds` set inside `ReconnectingPool`, capped at `MAX_KNOWN_IDS = 1000` and scoped to a single `subscribe()` call — it resets on every resubscribe and offers no protection across restarts, which is exactly when a retained/replayed backlog would be redelivered.

Separately, handlers registered via `NWCWalletServiceRequestHandler`only receive the *parsed* request body (e.g.`Nip47PayInvoiceRequest`), never the underlying event's `id`, `created_at`, or `tags`. This means a consuming application has no way to implement its own idempotency ledger or freshness check either — there's no stable identifier to key on, and no timestamp to bound how old a request is.

### Impact

For a wallet-service consumer (e.g. a Lightning wallet exposing `pay_invoice` over NWC), this combination means a captured/replayed
request can be re-executed with no live client and no valid secret needed beyond replaying previously-seen ciphertext to the wallet's
relay — including requests with a `created_at` far in the past. We reproduced this concretely against a consuming app (Lightning payment
re-dispatched after a simulated restart via relay replay; a previously failed payment succeeding on replay after a later restart; a 30-day-old
request being honored). Happy to share the PoC/repro if useful — didn't want to paste a large report inline here, but can add detail on request.

### Proposed change (open to a different shape — flagging the goal, not prescribing the API)

1. Accept an optional `since` (and maybe `until`) in the filter built by
   `subscribe()`, e.g. `subscribe(keypair, handler, { since })`, so
   consumers can bound replay of retained history using their own
   high-water mark. Doesn't stop a non-compliant relay from ignoring
   the filter, but helps well-behaved relays and reduces load either way.

[...]

PR #578, opened 2026-08-13
feat: add recordEvent handler to NWCWalletService for request deduplication

rolznz:
## Summary

Part of #577.

Adds an optional `recordEvent` method to `NWCWalletServiceRequestHandler`:

```ts
recordEvent?(
  eventId: string,
): NWCWalletServiceRecordEventResult | Promise<NWCWalletServiceRecordEventResult>;
```

It is called with the request event id after the event is decrypted and parsed, but before the request is handled. Returning `"ALREADY_PROCESSED"` makes the wallet service skip the request without responding (logged at debug level). This lets apps persist processed event ids (e.g. in their database) and deduplicate requests that a relay re-delivers — including across restarts, which the in-memory `knownIds` set in `ReconnectingPool` cannot cover.

`recordEvent` calls are serialized through a per-subscription promise queue, so duplicate events delivered in quick succession cannot pass an async check-then-record concurrently.

The `since`/`until` filter bounds from #577 are intentionally left for a separate PR.

## Response publish retries

While testing this end-to-end, the response event publish turned out to be a fire-and-forget `Promise.any` with no rejection handler — if all relays fail (or the pool is closed before the relay acks), it causes an unhandled promise rejection, which can crash a Node app. The publish now retries with exponential backoff (1s, 2s, 4s, 8s, 16s; 6 attempts total), logs each failed attempt, and stops retrying once the subscription is closed.

## Tests

- New unit tests (`src/nwc/NWCWalletService.test.ts`, first ones for `NWCWalletService`) run against a mocked pool with real signed + nip44-encrypted request events: normal handling, skipping on `"ALREADY_PROCESSED"`, serialization of concurrent duplicates, publish retry with backoff, and retry cancellation on unsubscribe.
- New e2e test (`e2e/nwc-wallet-service-record-event.test.ts`) runs a real `NWCWalletService` and `NWCClient` against relay.getalby.com and verifies the handler receives the request event id, and that a skipped request never reaches the handler and gets no response. (The relay-replays-old-events scenario itself can't be forced e2e since kind 23194 is in the ephemeral range.)

## Summary by CodeRabbit

* **Reliability**
  * Prevented previously processed wallet requests from being handled or answered again.
  * Serialized request processing to avoid duplicate handling during concurrent delivery.
  * Added automatic retries with increasing delays when publishing responses.
  * Stopped pending response retries when a subscription is closed.

* **Integration**
  * Added an optional event-recording hook so integrations can identify and skip requests they have already processed.

### DIFF
--- a/src/nwc/NWCWalletService.ts
+++ b/src/nwc/NWCWalletService.ts
@@ -34,6 +34,10 @@ export type NewNWCWalletServiceOptions = {
   logger?: Logger;
 };
 
+// First retry waits 1s; each subsequent attempt doubles (1s, 2s, 4s, 8s, 16s).
+const INITIAL_PUBLISH_RETRY_DELAY_MS = 1000;
+const MAX_PUBLISH_ATTEMPTS = 6;
+
 export class NWCWalletServiceKeyPair {
   walletSecret: string;
   walletPubkey: string;
@@ -100,6 +104,13 @@ export class NWCWalletService {
     await this._checkConnected();
 
     this.logger.debug("subscribing to relays");
+
+    // serializes handler.recordEvent calls so duplicate events delivered in
+    // quick succession cannot pass the check concurrently
+    let recordEventQueue: Promise<unknown> = Promise.resolve();
+
+    let unsubscribed = false;
+
     const sub = this.pool.subscribe(
       this.relayUrls,
 
@@ -126,6 +137,18 @@ export class NWCWalletService {
               method: Nip47Method;
               params: unknown;
             };
+
+            if (handler.recordEvent) {
+              const recordEventPromise = recordEventQueue.then(() =>
+                handler.recordEvent?.(event.id),
+              );
+              // keep the queue alive even if recordEvent throws
+              recordEventQueue = recordEventPromise.catch(() => {});
+              if ((await recordEventPromise) === "ALREADY_PROCESSED") {
+                this.logger.debug("skipped already processed event", event.id);
+                return;
+              }
+            }
 
             let responsePromise:
               | NWCWalletServiceResponsePromise<unknown>
@@ -206,8 +229,9 @@ export class NWCWalletService {
               keypair.walletSecret,
             );
 
-            // Try to publish to at least one relay
-            Promise.any(this.pool.publish(this.relayUrls, responseEvent));
+            // Try to publish to at least one relay, retrying with
+            // exponential backoff until the subscription is closed
+            this._publishWithRetry(responseEvent, () => unsubscribed);
           } catch (e) {
             console.error("Failed to handle event", e);
             return;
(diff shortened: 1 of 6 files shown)


src/nwc/NWCWalletServiceRequestHandler.ts
```diff
@@ -28,7 +28,21 @@ export type NWCWalletServiceResponsePromise<T> = Promise<{
...
+
 export interface NWCWalletServiceRequestHandler {
+  /**
+   * Called with the request event id after the event is decrypted and parsed,
+   * but before it is handled. Return "ALREADY_PROCESSED" to skip handling,
+   * e.g. if a relay re-delivers an event that was already processed.
+   * Calls are serialized by the wallet service, so duplicates delivered in
+   * quick succession will still be seen one at a time.
+   */
+  recordEvent?(
```
ajaysehwal, line 45: @rolznz I think we should clarify that `recordEvent` provides **event-level deduplication**, not payment idempotency. Since the event ID is recorded before the handler runs, a crash after recording but before handling can lose the request.  More importantly, NWC clients create a new event when retrying, so a new event ID can still trigger the same payment again. For `pay_invoice` / `pay_keysend` / `make_invoice`, we still need method-level idempotency (e.g., payment hash/invoice or an explicit idempotency key).

src/nwc/NWCWalletService.ts
```diff
@@ -206,8 +229,9 @@ export class NWCWalletService {
               keypair.walletSecret,
             );
 
-            // Try to publish to at least one relay
-            Promise.any(this.pool.publish(this.relayUrls, responseEvent));
+            // Try to publish to at least one relay, retrying with
+            // exponential backoff until the subscription is closed
```
ajaysehwal: @rolznz I think there’s a retry/shutdown issue here. The comment says publishing retries until the subscription is closed, but `close()` doesn’t set `unsubscribed`. So after `close()`, `_publishWithRetry()` can still retry `pool.publish()` several times on the closed pool.

I think the retry should also stop when the service is closed, and the comment should match the actual `MAX_PUBLISH_ATTEMPTS` behavior.

src/nwc/NWCWalletService.ts
```diff
@@ -100,6 +104,13 @@ export class NWCWalletService {
     await this._checkConnected();
 
     this.logger.debug("subscribing to relays");
+
+    // serializes handler.recordEvent calls so duplicate events delivered in
```
ajaysehwal, line 108: @rolznz, there’s a concurrency issue with `recordEvent` here. Every request waits for the previous `recordEvent` call, even when the event IDs are completely unrelated. A slow DB write for one event can block other NWC requests unnecessarily.
Could we make the event ID unique at the storage level and use that for deduplication instead? This would prevent the same event from being processed twice while allowing unrelated events to proceed concurrently.

src/nwc/NWCWalletService.ts
rolznz: Thanks, fixed

src/nwc/NWCWalletService.ts
rolznz: I don't think this is a real issue. The DB should be much faster than incoming NWC requests. What do you think? if your DB is that slow I think you will have bigger issues.

src/nwc/NWCWalletServiceRequestHandler.ts
rolznz: I added a comment about event-level deduplication, thanks.

I don't think we can do anything about keysend really, but you at the app level can decide I guess, if you want to rate limit keysend (if it's to the same destination with the same amount for example). For invoice payments your wallet/app layer should ensure that you do not pay the same invoice twice

rolznz:
@ajaysehwal I'll merge this for now. If you don't agree with any of my decisions let me know and we can make another issue to fix it :pray:

src/nwc/NWCWalletService.ts
ajaysehwal: Yeah, fair point. I agree it may not be a practical issue with a reasonably fast DB. My concern is mainly that we’re globally serializing unrelated events when we only really need atomic deduplication per event ID. But I’m fine with the current approach if we think the performance impact is negligible.

PR #580, opened 2026-08-13
fix: stop subscription reconnect loops when ReconnectingPool is closed

### LINKED ISSUE #579: ReconnectingPool: closing while a request is in flight leaks a reconnected socket
## Summary

Calling `close()` on an `NWCClient` (or `NWCWalletService`) while a subscription is still active does not stop that subscription's reconnect loop in `ReconnectingPool.subscribe()`. The loop observes the disconnect caused by `close()` and reconnects, opening a brand-new relay socket *after* the pool was closed. If the subscription is later closed (e.g. by a reply timeout), only the subscription is closed — the resurrected relay socket is never closed and leaks.

## Steps to reproduce

1. Create an `NWCClient` and call e.g. `getInfo()` against a wallet service that never responds (so the request's subscription stays open, waiting for its 10s reply timeout).
2. Call `client.close()` while the request is still pending.
3. Sequence that follows:
   - `pool.close()` closes the relay socket and removes it from the `relays` map.
   - The pending request's subscription sees the disconnect; its `runRelay` loop in `ReconnectingPool.subscribe()` waits ~1s and calls `ensureRelay()`, which creates and connects a **new** `Relay` because the old one was deleted from the map.
   - At 10s the reply timeout fires and calls `sub.close()` — this closes the subscription but not the relay, and nothing ever closes the new socket.
4. The process now holds an open TLS socket indefinitely (confirmed via `process._getActiveHandles()`: one `TLSSocket(readyState=open)` remains 12s+ after full cleanup).

This was discovered while writing the e2e test for #578: the jest process hung for minutes after all tests passed. The e2e test currently works around it by awaiting the client's reply timeout before closing (see the NOTE comment in `e2e/nwc-wallet-service-record-event.test.ts`).

## Impact

- Any app that closes a client/service with an in-flight request leaks a socket (and briefly runs a reconnect loop against a relay it asked to disconnect from).
- Long-lived processes accumulate leaked sockets; short-lived processes (tests, CLIs) hang on exit.

## Possible fixes

- Have `ReconnectingPool.close(relays)` / `destroy()` mark the pool (or the closed relay URLs) as closed so `runRelay` loops stop instead of reconnecting, and/or
- Track active subscriptions per pool and close them in `close()`/`destroy()`, so `NWCClient.close()` tears down in-flight request subscriptions too.

rolznz:
## Summary

Fixes #579.

Closing a `ReconnectingPool` (e.g. via `NWCClient.close()`) while a subscription was still active did not stop the subscription's reconnect loop: the loop observed the disconnect caused by `close()`, waited out its backoff, and re-opened a brand-new relay socket. Nothing ever closed that socket — leaked sockets in long-lived apps, hangs on exit in short-lived ones.

## Fix

Closing a pool is terminal (closing and re-using a pool was never supported), so the fix is a single pool-level `closed` flag:

- `close()` and `destroy()` set `closed = true` before closing the relay sockets.
- The reconnect loops in `subscribe()` check the flag at their existing exit points and stop instead of reconnecting. Loops mid-backoff exit when their timer fires; no new socket is ever opened.
- `ensureRelay()` throws on a closed pool, so no code path (subscribe loops, late publishes) can open a connection after close. The publish path already converts `ensureRelay` failures into a resolved "connection failure" string, so this surfaces gracefully.

Net change to `ReconnectingPool.ts`: +14/-5 lines.

## Tests

- New unit tests (`src/nwc/ReconnectingPool.test.ts`) mock `ensureRelay` (no network) and cover: close during reconnect backoff, close while connected, `destroy()`, closing the subscription itself, and the closed pool refusing new connections. The two close/destroy reconnect tests fail against the previous implementation.
- New e2e test (`e2e/nwc-close-in-flight-request.test.ts`) reproduces the original leak against relay.getalby.com: a client is closed while a `get_info` request is in flight (wallet service never responds), and 12s later no open TLS socket remains (verified via `process._getActiveHandles()`). Fails with 1 leaked socket on master, passes with this fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection shutdown behavior to prevent relay reconnections after the client, subscription, or wallet service is closed.
  * Closed connection pools now reject new relay connections and stop reconnect attempts.
  * Prevented lingering TLS sockets after closing during an in-flight request or reconnect backoff.

* **Tests**
  * Added comprehensive coverage for connection-pool lifecycle and shutdown scenarios.

### DIFF
--- a/e2e/nwc-close-in-flight-request.test.ts
+++ b/e2e/nwc-close-in-flight-request.test.ts
@@ -1,0 +1,59 @@
+import { generateSecretKey, getPublicKey } from "nostr-tools";
+import { bytesToHex } from "@noble/hashes/utils.js";
+import { NWCClient } from "../src/nwc/NWCClient";
+import {
+  NWCWalletService,
+  NWCWalletServiceKeyPair,
+} from "../src/nwc/NWCWalletService";
+
+const RELAY_URL = "wss://relay.getalby.com/v1";
+
+function getOpenTlsSockets(): number {
+  const handles = (
+    process as unknown as { _getActiveHandles: () => unknown[] }
+  )._getActiveHandles();
+  return handles.filter(
+    (h) =>
+      (h as object).constructor.name === "TLSSocket" &&
+      (h as { readyState?: string }).readyState === "open",
+  ).length;
+}
+
+test("closing a client with an in-flight request leaks no socket", async () => {
+  const walletSecret = bytesToHex(generateSecretKey());
+  const clientSecret = generateSecretKey();
+  const keypair = new NWCWalletServiceKeyPair(
+    walletSecret,
+    getPublicKey(clientSecret),
+  );
+
+  const service = new NWCWalletService({ relayUrls: [RELAY_URL] });
+  await service.publishWalletServiceInfoEvent(walletSecret, ["get_info"], []);
+  const unsubscribe = await service.subscribe(keypair, {
+    // never respond, so the client's request stays in flight
+    getInfo: () => new Promise(() => {}),
+  });
+
+  const client = new NWCClient({
+    nostrWalletConnectUrl: `nostr+walletconnect://${
+      keypair.walletPubkey
+    }?relay=${encodeURIComponent(RELAY_URL)}&secret=${bytesToHex(
+      clientSecret,
+    )}`,
+  });
+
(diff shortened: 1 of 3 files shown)


rolznz:
@coderabbitai review

PR #581, opened 2026-08-13
feat: add optional since/until filter to NWCWalletService.subscribe

### LINKED ISSUE #577: NWCWalletService.subscribe(): Unbounded replay of retained relay history

ajaysehwal:
## Summary
- Adds an optional third argument to `NWCWalletService.subscribe()` (`{ since?, until? }` in unix seconds) that is passed through to the NIP-01 kind 23194 request filter.
- Bounds are omitted unless set, so existing callers keep the previous unbounded subscription.
- This is a complementary bound for well-behaved relays that retain request history across resubscribe; `recordEvent` remains the idempotency path.

Closes #577

## Test plan
- [ ] Unit tests: `src/nwc/NWCWalletService.test.ts` (omit both bounds, include both, include `since` only)
- [ ] Confirm existing `subscribe(keypair, handler)` callers still type-check and subscribe without `since`/`until`
- [ ] Optional: subscribe with `since` set to a recent high-water mark against a retaining relay and confirm older events are not delivered

## Summary by CodeRabbit

- **New Features**
  - Added optional `since` and `until` time filters to wallet service subscriptions.
  - Supports limiting replayed events during initial subscriptions and resubscriptions.

- **Documentation**
  - Updated the wallet service quick-start guide with filter usage, Unix-second formatting, and persisted high-water mark behavior.

- **Tests**
  - Added coverage for default, complete, and partially specified subscription filters.

### DIFF
--- a/docs/nwc-wallet-service.md
+++ b/docs/nwc-wallet-service.md
@@ -52,3 +52,5 @@ const unsub = await walletService.subscribe(keypair, {
   // ... handle other NIP-47 methods here
 });
 ```
+
+`subscribe()` accepts an optional third argument `{ since?, until? }` (unix seconds) that is passed through to the NIP-01 request filter. Use a persisted high-water mark as `since` to limit replay of retained history on resubscribe. Relays can ignore these bounds, so keep using `recordEvent` for idempotency. 
--- a/src/nwc/NWCWalletService.test.ts
+++ b/src/nwc/NWCWalletService.test.ts
@@ -1,6 +1,10 @@
-import { generateSecretKey, getPublicKey, Event } from "nostr-tools";
+import { generateSecretKey, getPublicKey, Event, Filter } from "nostr-tools";
 import { bytesToHex } from "@noble/hashes/utils.js";
-import { NWCWalletService, NWCWalletServiceKeyPair } from "./NWCWalletService";
+import {
+  NWCWalletService,
+  NWCWalletServiceKeyPair,
+  NWCWalletServiceSubscribeFilter,
+} from "./NWCWalletService";
 import { NWCWalletServiceRequestHandler } from "./NWCWalletServiceRequestHandler";
 import { Nip47GetInfoResponse } from "./types";
 
@@ -10,6 +14,7 @@ type SubscribeParams = {
 
 async function setupSubscribedWalletService(
   handler: NWCWalletServiceRequestHandler,
+  filter?: NWCWalletServiceSubscribeFilter,
 ) {
   const walletSecret = bytesToHex(generateSecretKey());
   const clientSecret = generateSecretKey();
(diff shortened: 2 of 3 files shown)


rolznz (approved):
utACK

PR #585, opened 2026-09-07
fix: move @webbtc/webln-types to dependencies

### LINKED ISSUE #234: Build error caused by @webbtc/webln-types
The error I get: `Cannot find module '@webbtc/webln-types' or its corresponding type declarations.`

@webbtc/webln-types is declared as dev dependency but imported in the code, is this why?

Steve0x2a:
Published `.d.ts` files import `@webbtc/webln-types`, but that package was only listed under `devDependencies`. TypeScript consumers that do not install it themselves fail with:

```
Cannot find module '@webbtc/webln-types' or its corresponding type declarations.
```

`import type` is used at the source import sites. The published declarations still re-export those types, so the package is moved to `dependencies` so package managers install it for consumers.

A smoke test packs the SDK and typechecks a clean TypeScript project that does not add `@webbtc/webln-types` itself.

Fixes #234

Reported by @rolznz

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured published packages include the WebLN type definitions required for consumers to typecheck successfully without additional manual installation.

* **Tests**
  * Added coverage verifying clean, production-only installations can compile consumer code using the SDK and OAuth APIs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
   },
   "dependencies": {
     "@getalby/lightning-tools": "^8.1.1",
+    "@webbtc/webln-types": "^3.0.0",
     "nostr-tools": "^2.23.3"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


rolznz:
Hi, thanks for the PR! I'll review and test this before doing the next release.

